PR #14, opened 2018-06-27
Task/10/verify signature

aleixmorgadas:
@aizaiz, you might be interested in the changes I made.

- Added the method `signData` to `Account` class
- Changed `verify` method from static to instance method of `PublicAccount`

### DIFF
--- a/src/model/account/PublicAccount.ts
+++ b/src/model/account/PublicAccount.ts
@@ -57,14 +57,13 @@ export class PublicAccount {
     /**
      * Verify a signature.
      *
-     * @param {PublicAccount} publicAccount - The public account to use for verification.
      * @param {string} data - The data to verify.
      * @param {string} signature - The signature to verify.
      *
      * @return {boolean}  - True if the signature is valid, false otherwise.
      */
-    static verifySignature(publicAccount: PublicAccount, data: string, signature: string): boolean {
-        if (!publicAccount || !data || !signature) {
+    public verifySignature(data: string, signature: string): boolean {
+        if (!data || !signature) {
             throw new Error('Missing argument');
         }
 
@@ -89,7 +88,7 @@ export class PublicAccount {
         // Convert to Uint8Array
         convertedData = convert.hexToUint8(convertedData);
 
-        return KeyPair.verify(convert.hexToUint8(publicAccount.publicKey), convertedData, convertedSignature);
+        return KeyPair.verify(convert.hexToUint8(this.publicKey), convertedData, convertedSignature);
     }
 
     /**
--- a/src/model/account/PublicAccount.ts
+++ b/src/model/account/PublicAccount.ts
@@ -80,15 +80,8 @@ export class PublicAccount {
         // Convert signature key to Uint8Array
         const convertedSignature = convert.hexToUint8(signature);
 
-        let convertedData;
-
-        // Convert data to hex if data is not hex
-        if (!convert.isHexString(data)) {
-            convertedData = convert.utf8ToHex(data);
-        }
-
         // Convert to Uint8Array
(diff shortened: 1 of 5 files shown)


src/model/account/PublicAccount.ts
```diff
@@ -53,6 +54,43 @@ export class PublicAccount {
...
+    /**
+     * Verify a signature.
+     *
+     * @param {string} data - The data to verify.
+     * @param {string} signature - The signature to verify.
+     *
+     * @return {boolean}  - True if the signature is valid, false otherwise.
+     */
+    public verifySignature(data: string, signature: string): boolean {
+        if (!data || !signature) {
```
Vektrat: is it possible there is a case where the signed data is ""? (no data)

src/model/account/PublicAccount.ts
```diff
@@ -53,6 +54,43 @@ export class PublicAccount {
...
+            throw new Error('Missing argument');
+        }
+
+        if (signature.length !== 128) {
```
Vektrat: can we use a reference to some declaration of what this size actually is?

src/model/account/PublicAccount.ts
```diff
@@ -53,6 +54,43 @@ export class PublicAccount {
...
+        if (!convert.isHexString(signature)) {
+            throw new Error('Signature must be hexadecimal only');
+        }
+
+        // Convert signature key to Uint8Array
+        const convertedSignature = convert.hexToUint8(signature);
+
+        let convertedData;
+
+        // Convert data to hex if data is not hex
```
Vektrat: I made this comment in the previous specific PR, are we sure we want to dynamically parse types in this function? Also, maybe the function documentation should specify the expected data format

test/model/account/Account.spec.ts
```diff
@@ -45,4 +46,14 @@ describe('Account', () => {
...
 
+    it('signData', () => {
+        const account = Account.createFromPrivateKey(
+            'AB860ED1FE7C91C02F79C02225DAC708D7BD13369877C1F59E678CC587658C47',
+            NetworkType.MIJIN_TEST,
+        );
+        const publicAccount = account.publicAccount;
+        const signed = account.signData('catapult rocks!');
+        expect(publicAccount.verifySignature('catapult rocks!', signed))
+            .to.be.equal(true);
```
Vektrat: `to.be.true`

I've seen this repeats in other tests as well ^^

test/model/account/PublicAccount.spec.ts
```diff
@@ -27,3 +27,84 @@ describe('PublicAccount', () => {
...
+
+        // Act & Assert:
+        expect(signerPublicAccount.verifySignature(data, signature)).equal(false);
+    });
+
+    it('Throw error if signature verification is missing a parameter', () => {
+        // Arrange:
+        const signerPublicAccount = PublicAccount.createFromPublicKey('22816F825B4CACEA334723D51297D8582332D8B875A5829908AAE85831ABB508',
+            NetworkType.MIJIN_TEST);
+        const data = '';
```
Vektrat: as commented before, is this an invalid case?

src/model/account/PublicAccount.ts
aleixmorgadas: Yes, does it make sense? Or should we thrown an exception?

test/model/account/Account.spec.ts
aleixmorgadas: `to.be.true` is an issue with tslint, we should 

a) change the tslint rules
b) use `equal(true)`.

I like `to.be.true`

Vektrat: I prefer to accept `to.be.true` as well, but np it's a detail

PR #18, opened 2018-07-11
Exposed Deadline to the public in Transaction class

gimyboya:


gimyboya:
implementation and test done. please review @aleixmorgadas

package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "pretest": "npm run build",
     "test": "mocha --ui bdd --recursive ./dist/test --timeout 90000",
+    "tran": "mocha --ui bdd --recursive ./dist/test/model/transaction/Transaction.spec --timeout 90000",
```
aleixmorgadas: why?

src/model/transaction/Transaction.ts
```diff
@@ -156,4 +156,18 @@ export abstract class Transaction {
...
         return parseInt(versionDTO, 16);
     }
+
+    /**
+     * @description re-aplly a given value to the transaction in an immutable way
+     * @param {Deadline} deadline
+     * @returns {Transaction}
+     * @memberof Transaction
+     */
+    public reapplygiven(deadline: Deadline): Transaction {
```
aleixmorgadas: replace `reapplygiven` by `reply` or `replyGiven`

src/model/transaction/Transaction.ts
```diff
@@ -156,4 +156,18 @@ export abstract class Transaction {
...
+        if (this.isUnannounced()) {
+            return Object.assign({__proto__: Object.getPrototypeOf(this)}, this, {deadline});
+        } else {
+            throw new Error('an Announced transaction can\'t be modified');
```
aleixmorgadas: does exist a case where you want to reply an already announced transaction?

I would do:
```typescript
if (this.isUnannounced()) {
    return Object.assign({__proto__: Object.getPrototypeOf(this)}, this, {deadline});
}
throw new Error('an Announced transaction can\'t be modified');
```

The code is more clean, personal style :smile:

test/model/transaction/Transaction.spec.ts
```diff
@@ -98,6 +98,52 @@ describe('Transaction', () => {
             expect(transaction.hasMissingSignatures()).to.be.equal(true);
         });
     });
+
+    describe('reapplygiven', () => {
```
aleixmorgadas: updated according to the previous comments

src/model/transaction/Transaction.ts
aleixmorgadas: I would add the deadline as optional parameter, like

```typescript
public reply(deadline: Deadline = Deadline.create()): Transaction
```

test/model/transaction/Transaction.spec.ts
```diff
@@ -98,6 +98,52 @@ describe('Transaction', () => {
...
+            const transaction = new FakeTransaction(TransactionType.TRANSFER,
+                NetworkType.MIJIN_TEST,
+                1,
+                Deadline.create(),
+                UInt64.fromUint(0),
+                undefined,
+                undefined,
+            );
+
+            const newTransaction = transaction.reapplygiven(Deadline.create(3));
```
aleixmorgadas: I would move check that the new deadline is assigned, something like:

```typescript
const newDeadline = Deadline.create(3);
const newTransaction = transaction.reapplygiven(newDeadline); 
expect(newTransaction.deadline).to.be.equal(newDeadline);
```

package.json
gimyboya: sorry, I was just testing and forgot to remove it :P it's faster because somehow in my console mocha couldn't filter this `./src/test/model/transaction/Transaction.spec`

PR #19, opened 2018-08-14
adding unsubscription methods

aleixmorgadas:
:warning: work in progress

Listener provides the option to stop the subscription to different channels.


src/infrastructure/Listener.ts
```diff
@@ -320,6 +320,18 @@ export class Listener {
         this.webSocket.send(JSON.stringify(subscriptionMessage));
     }
 
+    /**
+     * @internal
+     * @param channel - Channel to unsubscribe
+     */
+    private unsubscribeTo(channel: string) {
```
evias: this should be renamed to `unsubscribeFrom`

evias: renamed in [#559af45](https://github.com/evias/nem2-sdk-typescript-javascript/commit/559af45c598b7742952f4ca54bebde679996680c)

Issue #36, opened 2018-12-13
UInt64.fromUInt does not throw in case of negative values


evias:
1. Expected behavior and actual behavior.

When using the `UInt64.fromUInt` method, there is no error handling done when `negative values` are used to create the instance. As a result, **because it does not produce an error**, requests to the API are sent with `4294867296` amounts when you initialize them with `UInt64.fromUint(-1)`. 

The expected behaviour would be to **throw an Error** in case of malformed Unsigned Integer amounts as this would avoid such collaterals.

2. Steps to reproduce the problem.

This example is not from me, but will reproduce the problem at hand:

```
MosaicSupplyChangeTransaction.create(
    Deadline.create(),
    mosaicID,
    MosaicSupplyType.Decrease,
    UInt64.fromUint(-1),
    NetworkType.MIJIN_TEST);
```

3. Specifications like the version of the project, operating system, or hardware.

nem2-sdk@0.10.1

PR #37, opened 2018-12-13
nemtech/nem2-sdk-typescript-javascript#36 : Added error thrown for UI…

evias:
…nt64.fromUInt() calls with negative values; Added negative amount failure test case for fromUint()

### DIFF
--- a/src/model/UInt64.ts
+++ b/src/model/UInt64.ts
@@ -37,6 +37,10 @@ export class UInt64 {
      * @returns {UInt64}
      */
     public static fromUint(value: number): UInt64 {
+        const uintArray = [(value & 0xFFFFFFFF) >>> 0, (value / 0x100000000) >>> 0];
+        if (0x00200000 <= uintArray[1] || 0 > value || 0 !== (value % 1))
+            throw Error(`number cannot be converted to uint '${value}'`);
+
         return new UInt64(uint64.fromUint(value));
     }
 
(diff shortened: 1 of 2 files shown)


evias:
fix for nemtech/nem2-sdk-typescript-javascript#13

evias:
Tests were also commented out.

Transactions can currently be *created* with `null` values when sent to the API. It is a bugfix that allows for more stability and I believe any edge case that is not handled by this fix, currently has lesser importance.

The code duplication did hit me of course but people using the nem2-sdk should be receiving an error no matter if they use fromUInt or the constructor, and it was not the case before. 

Additionally, whether fromUInt will still be using nem2-library in the future is probably worth a question and for that matter the important thing is you can't create UInt objects with negative values also when using the nem2-sdk UInt64 objects.

I decided to put the fix in here because in this sdk, there were also no unit tests done for the `fromUInt` method with negative values passed to it, yet the UInt64 class can be used to create UInt64 objects.. Probably the SDK's UInt64 objects can be upgraded at some point forward. 

Thanks for the feedback 👍

evias:
I checked a previous version of this source code now (nem2-library), back from a few months ago.

I see a commit from Guillem, with message: `CHANGE fromUint don't check values when creating uint array`. Also, back then there didnt seem to be test cases for the uint64 class in nem2-library.

From this I believe that it was removed before the introduction of unit tests. And the unit tests were implemented and commented out to be enabled later.

Issue #29, opened 2018-11-24
Argument order of BlockchainStorageInfo constructor is shifted


planethouki:
## 1. Expected behavior and actual behavior.

BlockchainStorageInfo obtained by blockchainHttp.getDiagnosticStorage() has invalid key-value pair.

`numAccounts` shows number of blocks
`numBlocks` shows number of transactions
`numTransactions` shows number of accounts

### Expected behavior

```javascript
BlockchainStorageInfo { 
	numBlocks: 89263, 
	numTransactions: 444, 
	numAccounts: 81 
}
```

### Actual behavior

```javascript
BlockchainStorageInfo { 
	numAccounts: 89263, 
	numBlocks: 444, 
	numTransactions: 81 
}
```

To compare, rest api of `/diagnostic/storage` returns
```json
{
	"numBlocks": 89263,
	"numTransactions": 444,
	"numAccounts": 81
}
```

## 2. Steps to reproduce the problem.

Run this code.

```javascript
const nem2Sdk = require("nem2-sdk");
const BlockchainHttp = nem2Sdk.BlockchainHttp;
const host = <something host>;
const blockchainHttp = new BlockchainHttp(host);
blockchainHttp.getDiagnosticStorage().subscribe(
    x => console.log(x)
);
```

## 3. Specifications like the version of the project, operating system, or hardware.

node: v10.11.0
nem2-sdk: 0.10.1
rest-gateway is docker image: techbureau/catapult-rest-server:1.0.7
catapult node is docker image: techbureau/catapult-server:gcc-0.1.0.2

## 4. More information

https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/master/src/model/blockchain/BlockchainStorageInfo.ts
```javascript
constructor(
	public readonly numAccounts: number,
	public readonly numBlocks: number,
	public readonly numTransactions: number) {
}
```

is should be

```javascript
constructor(
	public readonly numBlocks: number,
	public readonly numTransactions: number,
	public readonly numAccounts: number) {
}
```

and also test

https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/master/test/model/blockchain/BlockchainStorageInfo.spec.ts

evias:
Since the wrong parameters order was only defined in the *model* constructor and everywhere else several calls to the BlockchainStorageInfo class would imply the aforementioned `Expected Behaviour`, it seemed like an important fix to make.

[...]

PR #40, opened 2019-01-14
Change the order of BlockchainStorageInfo arguments

44uk:
`http://localhost:3001/diagnostic/storage` returns response like below.

```
{
numBlocks: 1076,
numTransactions: 31,
numAccounts: 26
}
```

but `BlockchainHttp#getDiagnosticStorage` returns below.

```
const nem2Sdk = require("nem2-sdk");
const BlockchainHttp = nem2Sdk.BlockchainHttp;
const blockchainHttp = new BlockchainHttp('http://localhost:3000');
blockchainHttp
    .getDiagnosticStorage()
    .subscribe(block => console.log(block), err => console.error(err));
# output
# BlockchainStorageInfo { numAccounts: 1076, numBlocks: 31, numTransactions: 26 }
```

### DIFF
--- a/src/infrastructure/BlockchainHttp.ts
+++ b/src/infrastructure/BlockchainHttp.ts
@@ -154,9 +154,9 @@ export class BlockchainHttp extends Http implements BlockchainRepository {
         return observableFrom(
             this.blockchainRoutesApi.getDiagnosticStorage()).pipe(map((blockchainStorageInfoDTO) => {
             return new BlockchainStorageInfo(
+                blockchainStorageInfoDTO.numAccounts,
                 blockchainStorageInfoDTO.numBlocks,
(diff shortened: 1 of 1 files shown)


evias:
As to stay compatible with the current catapult-rest implementation, I would rather want to fix the model itself + the unit test to be compatible with following format : https://github.com/nemtech/catapult-rest/blob/master/rest/src/db/CatapultDb.js#L187

```
    numBlocks: xxxx
    numTransactions: xxxx
    numAccounts: xxxx
```

I did a commit in https://github.com/nemtech/nem2-sdk-typescript-javascript/issues/29 to fix this issue - @dgarcia360 what do you think ?

evias:
fixed in PR #63

PR #43, opened 2019-02-03
Issue #36, Error handling on Uint64.fromUnit with negative value.

rg911:
Couple of tslint issues fixed on MosaicProperty as well

### DIFF
--- a/src/model/UInt64.ts
+++ b/src/model/UInt64.ts
@@ -38,7 +38,7 @@ export class UInt64 {
      */
     public static fromUint(value: number): UInt64 {
         if (value < 0) {
-            throw new Error('Unit value cannot be negative');
+            throw new Error('Unsigned integer cannot be negative');
         }
         return new UInt64(uint64.fromUint(value));
     }
--- a/src/model/UInt64.ts
+++ b/src/model/UInt64.ts
@@ -37,6 +37,9 @@ export class UInt64 {
      * @returns {UInt64}
      */
     public static fromUint(value: number): UInt64 {
+        if (value < 0) {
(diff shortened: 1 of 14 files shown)


src/model/UInt64.ts
```diff
@@ -37,6 +37,9 @@ export class UInt64 {
      * @returns {UInt64}
      */
     public static fromUint(value: number): UInt64 {
+        if (value < 0) {
```
evias: This check would better be done in the class constructor.

Also the error message contains a typo. Either of 'UInt' or 'Unsigned Integer' would be fine 👍

test/model/UInt64.spec.ts
```diff
@@ -64,6 +64,12 @@ describe('Uint64', () => {
         expect(uint64Compact).to.be.equal(51110867862);
     });
 
+    it('should fromUnit throw exception with negative unit value', () => {
```
evias: Same typo in UInt.

src/model/UInt64.ts
rg911: @evias 32bit unsigned int capped at 4294967295 and -1 >>> 0 === 4294967295. Constructor only takes UInt arrays, how does it differentiate -1 and real 4294967295?

evias: I agree @rg911, you are right. 
I believe adding it in fromUInt makes sense like this then :+1:

rg911: 👍

rg911: More clean way is to add the check in NEM2-library I'd thought?

Issue #24, opened 2018-09-25
getTransactionStatus fails when querying for failed transactions


jorisadri:
When querying for transaction status and the transaction failed. The <a href="https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/master/src/infrastructure/TransactionHttp.ts#L96">height</a> is not provided by the api but it is expected in the SDK

jorisadri:
Sorry I was wrong, it does not fail when the height is not provided.

rg911:
https://github.com/nemtech/nem2-sdk-typescript-javascript/pull/59
Fixed, unit test to be added after nem2-lib decommissioned.

evias:
Fixed with PR #59

PR #59, opened 2019-02-13
Fix getTransactionStatus for failed transaction (#24)

rg911:
Fixed the issue but unit testing on hold after nem2-lib decommissioned.


test/model/UInt64.spec.ts
```diff
@@ -64,7 +64,7 @@ describe('Uint64', () => {
         expect(uint64Compact).to.be.equal(51110867862);
     });
 
-    it('should fromUnit throw exception with negative unit value', () => {
+    it('should fromUnit throw exception with negative uint value', () => {
```
evias: @rg911 please also fix the `fromUnit` typo in this message

Issue #58, opened 2019-02-11
New lock hash algorithms for secret lock


dgarcia360:
Add new lock hash algorithms for secret lock transaction https://github.com/nemtech/catapult-server/issues/13

* [HashType](https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/6f71e1f4f88f995b94308da3d2c45ff31b5fccef/src/model/transaction/HashType.ts)

rg911:
This one is with me now :D

PR #62, opened 2019-02-15
Issue/#57 #58 secretlock sha512 to sha256 && New lock hash algorithms for secret lock

rg911:
New hash types added based on [newLockHashAlgorithm](https://github.com/nemtech/catapult-server/blob/master/plugins/txes/lock_secret/src/model/LockHashAlgorithm.h) 

Changed default secret lock hash type SHA3_512 to SHA3_256

tests added


rg911:
will create a new one remove all unmerged branches

Issue #55, opened 2019-02-11
Alias


dgarcia360:
### Set/remove an alias to an address

```
AliasTransaction.createForAddress(
    Deadline.create(),
    AliasAction.Add, //AliasAction.Add or AliasAction.Remove
    namespaceId,
    address,
    NetworkType.MIJIN_TEST
);
```

### Set/remove an alias to a mosaic

```
AliasTransaction.createForMosaic(
    Deadline.create(),
    AliasAction.Add, //AliasAction.Add or AliasAction.Remove
    namespaceId,
    mosaicID,
    NetworkType.MIJIN_TEST
);
```

To define:

* Transfer a mosaic (alias)
* Transfer to an account (alias)
* Get address by alias
* Get mosaicId by alias
* Get mosaic alias
* Get address alias

jorisadri:
A suggestion for the "To define" functions. 

### Get address by alias
```
NamespaceHttp.getLinkedAddress(alias);
  // returns address
```
### Get mosaicId by alias
```
NamespaceHttp.getLinkedmosaicId(alias);
  // returns mosaicId
```
###  Get address alias
```
AccountHttp.getLinkedAlias(address/publicKey);
  // returns namespace
```
###  Get mosaic alias
```
MosaicHttp.getLinkedAlias(mosaicId);
  // returns namespace
```

dgarcia360:
@jorisadri +1 to the suggested functions. Since there are not specific endpoints for alias, seems a good idea reuse mosaic/namespace http instead of creating the ``AliasHttp``.

evias:
Still missing *Http implementations.

evias:
Missing tests for NamespaceHttp.getLinkedMosaicId and getLinkedAddress.

Also missing AccountHttp and MosaicHttp implementation.

evias:
Fixed in PR #66

evias:
merged in v0.10.3-1

dgarcia360:
I open the issue again since ``transfer to an account (alias)`` is not possible yet. 

The recipient field should be an address or an alias (namespaceId) if possible.

evias:
PR #83 created with implementation proposals for aliased recipients. :v:

PR #83, opened 2019-03-12
#55 : Added Alias capability for TransferTransaction.recipient field …

evias:
Implementation proposal for issue #55  : Added Alias capability for TransferTransaction.recipient field with `Address | NamespaceId` type.

- Includes an update of `Listener` class to detect `signer` and `recipient` fields.
- Includes an update of `TransferTransaction` class to add `recipientToString()` helper.

### DIFF
--- a/src/model/transaction/TransferTransaction.ts
+++ b/src/model/transaction/TransferTransaction.ts
@@ -94,7 +94,7 @@ export class TransferTransaction extends Transaction {
      * @internal
      * @returns {string}
      */
-    protected recipientToString(): string {
+    public recipientToString(): string {
         if (this.recipient instanceof NamespaceId) {
             // namespaceId available, return hexadecimal notation
             return (this.recipient as NamespaceId).toHex();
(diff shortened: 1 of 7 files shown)


src/model/transaction/TransferTransaction.ts
```diff
@@ -88,6 +89,21 @@ export class TransferTransaction extends Transaction {
         super(TransactionType.TRANSFER, networkType, version, deadline, fee, signature, signer, transactionInfo);
     }
 
+    /**
+     * Return the string notation for the set recipient
+     * @internal
+     * @returns {string}
+     */
+    protected recipientToString(): string {
```
rg911: test on this new method?

evias: added in [#00938be](https://github.com/nemtech/nem2-sdk-typescript-javascript/pull/83/commits/00938bee4f125e7e0b10aedb3cc5ab4aea8fd7e8) :+1:

Issue #56, opened 2019-02-11
Expose serialization methods


dgarcia360:
The NEM-SDK has a class [TransactionMapping ](https://github.com/nemtech/nem2-sdk-java/blob/450676210c5a75f4624fc3bee7bbde96740109d1/src/main/java/io/nem/sdk/infrastructure/TransactionMapping.java) to map JSON transactions into models.

It can be [useful](https://github.com/nemtech/NIP/issues/6) for advanced users to make use the json mapping, and as well get transactions in serialized form. This table gathers the functions that each transaction model could implement.

<table>
  <tr>
   <td><strong>Function</strong>
   </td>
   <td><strong>Description</strong>
   </td>
  </tr>
  <tr>
   <td>serialize():string
   </td>
   <td>Serializes a transaction model into binary (unsigned payload)
   </td>
  </tr>
  <tr>
   <td>toDTO():json	
   </td>
   <td>Serializes a transaction model into JSON
   </td>
  </tr>
  <tr>
   <td>createFromPayload(data:string):Transaction
   </td>
   <td>Reads the unsigned payload and creates a Transaction.
   </td>
  </tr>
  <tr>
   <td>createFromDTO(data:json):Transaction
   </td>
   <td>Reads from JSON and creates a Transaction.
   </td>
  </tr>
</table>

dgarcia360:
@rg911 Have renamed the name of the functionsto achieve more consistency.

evias:
Merged in master, to be published in v0.11.2

aleixmorgadas:
Hi, [nem2-wallet-browserextension](https://github.com/aleixmorgadas/nem2-wallet-browserextension) has a heavy dependency on this feature, and it's about to be blocked the development since we are reaching the point that we are working with the integration between components.

What prevents to publish `v0.11.2`?

rg911:
> Hi, [nem2-wallet-browserextension](https://github.com/aleixmorgadas/nem2-wallet-browserextension) has a heavy dependency on this feature, and it's about to be blocked the development since we are reaching the point that we are working with the integration between components.
> 
> What prevents to publish `v0.11.2`?

Hi mate, nothing specific, it's all merged. we have got a few tickets done under testing and would like to wrapt them in as well. Let us know if it's blocking your guys.

aleixmorgadas:
I mean, yes, it's blocking us because we need those methods to make it simpler. I don't need that tomorrow, but probably next week.

[...]

PR #86, opened 2019-03-13
Task/g56 expose serialization methods

rg911:
Issue #56 

Added:

- ### Utility/TransactionMapping

1. createFromDTO(dataJson: object): Transaction 
2. createFromPayload(dataBytes: string): Transaction
    
- ### Transaction - new methods
1. Transaction.toJSON(): string - create JSON object from Transaction class.
2. Transaction.serialize(): string - serialize Transaction (unsigned payload)


rg911:
@evias  I managed to fixed the bug in RegisterNamespaceTransaction now.

All done!

src/model/account/Address.ts
```diff
@@ -113,4 +113,14 @@ export class Address {
     public equals(address: Address): boolean {
         return this.plain() === address.plain() && this.networkType === address.networkType;
     }
+
+    /**
+     * Create DTO object
+     */
+    public toDTO() {
```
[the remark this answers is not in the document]
evias: @rg911  ^

src/model/transaction/AggregateTransactionCosignature.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 import {PublicAccount} from '../account/PublicAccount';
-
+import { NetworkType } from '../blockchain/NetworkType';
```
evias: why include NetworkType in this file ?

src/model/transaction/RegisterNamespaceTransaction.ts
```diff
@@ -79,7 +79,9 @@ export class RegisterNamespaceTransaction extends Transaction {
             new UInt64([0, 0]),
             NamespaceType.SubNamespace,
             namespaceName,
-            new NamespaceId(subnamespaceNamespaceId(parentNamespace, namespaceName)),
+            typeof parentNamespace === 'string' ?
```
evias: you are using *type definition* as the way to determine whether you should be creating a subnamespace id or a namespace id ? I believe it would be best to to handle this with the namespace.type content.

src/utility/TransactionMapping.ts
```diff
@@ -0,0 +1,40 @@
+/*
```
evias: folder hierarchy should be consistent with the [nemtech/nem2-sdk-java](https://github.com/nemtech/nem2-sdk-java).

better would be to introduce the `utility` folder as `src/core/utils` and have `TransactionMapping` added in there

src/model/transaction/AggregateTransactionCosignature.ts
rg911: It's a good question, some how, it's not directly referenced in the class, but when trying to call the method, it keeps complaining missing this ref.  I will ave a look anyway, could be VS code's issue

src/utility/TransactionMapping.ts
rg911: sure, will change it

src/model/transaction/RegisterNamespaceTransaction.ts
rg911: This is not to determine the type of the namespace, it determine the type of the input whether a string (name) or uint (number[]). if a number[] passed in, we don't need to call 'subnamespaceNamespaceId' to convert the string to number[].

Issue #97, opened 2019-03-19
CreateTransactionFromDTO - MosaicAlias not mapping properly


rg911:
1. aliasAction property is wrong. should be *aliasAction* instead of *actionType*
2. mosaicId is parse from raw Uint64 should create instance of MosaicId instead
3. namespaceId, same issue as mosaicId

** similar issue on AddressAliasTransaction

rg911:
I will be taking this one

evias:
merged in master

will be integrated in later to be published v0.11.2

PR #98, opened 2019-03-19
Added #97  CreateTransactionFromDTO - MoasicAlias

rg911:
Issue: #97 

fixed bug in CreateTransactionFromDTO - MoasicAlias


rg911:
Also Added address alias fixes

Issue #100, opened 2019-03-25
BlockInfo class upgrade for cow blocks


evias:
1. Expected behavior and actual behavior.

Following fields are missing in the BlockInfo class:

- `feeMultiplier`
- `blockReceiptsHash`
- `stateHash`
- `beneficiaryPublicKey`

2. Steps to reproduce the problem.

use `BlockchainHttp.getBlockByHeight()`

evias:
Opened PR #101 with bugfix for this issue.

evias:
merged in [v0.11.3](https://github.com/nemtech/nem2-sdk-typescript-javascript/releases/tag/v0.11.3)

Issue #53, opened 2019-02-11
Fees


dgarcia360:
https://nemtech.github.io/concepts/transaction.html#fees

Include ``max_fee`` field in the transaction.

```
TransferTransaction.create(
    Deadline.create(),  
    Address.createFromRawAddress('S..4'),
    [new Mosaic( Alias.create('alice.token'), UInt64.fromUint(10)],
    PlainMessage.create('Welcome To NEM'),
    NetworkType.MIJIN_TEST,
    NetworkCurrencyMosaic.createRelative(10), // maxFee);
```

dgarcia360:
From Slack:

1. Replace `fee` for `max_fee` and make this parameter settable from the transaction model.

⚠️ If this could be an optional parameter, it won't break the current guides. Default to 0 if not set.

2. As `effective_fee = transaction::size * block::fee_multiplier`, it could be useful to have a method to calculate which is the transaction size, so that max_fee parameter can be set accordingly.

3. `TransactionHttp` (or another service) could provide with a function to calculate the fee paid given the transaction hash.

evias:
Now maxFee field is left optional from all `create()` methods.

Also, the fee field has been renamed `maxFee` on all transaction constructors and unit tests were adapted.

Still missing: 

- transaction size calculator
- transaction paid reader

evias:
Added `Transaction.size` getter to retrieve base transaction size.

Notes:

- Currently the `signature` and `signer` field are assumed to be set.
- Each class extending the Transaction class implemented an overload of the `size` getter to define their type-specific binary data sizes.

This is an implementation proposal for the `size` feature to avoid adding a `switch(this.type)` with many transaction type cases, and make use classes inheritance instead.

Unit tests have been added for each `size` call.

evias:
Open PR #101 with implementation proposal to be merged.

PR #101, opened 2019-03-25
Implementation proposal for task/g53 : Fees

evias:
Implementation proposal for Fees features including:

### Added 

- Renamed `fee` transaction classes fields to `maxFee` (#53 )
- Added `maxFee` *optional* field to `create()` factories. (#53 )
- Added `Transaction.size` getter method to return base transaction byte size (#53 )
  - The calculation currently assumes `signature` and `signer` fields to be set.
- Added `size` getter overload using `Transaction.size` base getter in all transaction type classes  (#53 )
  - Each transaction type defines its own type specific data byte size.
- Added `TransactionHttp.getTransactionEffectiveFee(transactionHash|transactionId)` (#53 )

### Fixed

- Fixed `BlockInfo` model for missing fields (#100 )


evias:
Merge stalled - Problem with BlockInfo upgrade with `beneficiaryPublicKey` field, breaking most of the transaction listeners.

Awaiting more details.

PR #108, opened 2019-04-02
Fixed #107 - Fixed issues on createFromDTO with JSON payload

### LINKED ISSUE #107: TransactionMapping fromDTO fails to load transactions

Using ``TransactionMapping`` class we can create Transactions using the DTO format. Once we turn a transaction to DTO, and try to recreate it using  ``TransactionMapping.fromDTO(transaction.toJSON())`` we get the following error:

> TypeError: Cannot read property 'type' of undefined from \nem2sdk\dist\src\infrastructure\transaction\CreateTransactionFromDTO.js:60:36.

``` typescript
    const transferTransaction = TransferTransaction.create(
        Deadline.create(),
        Address.createFromRawAddress('SAGYCE-QM5SK2-TGFUC5-Z5GZJR-ATKTBS-UQQMMH-KW5B'),
        [NetworkCurrencyMosaic.createRelative(10)],
        PlainMessage.create('hello'),
        NetworkType.MIJIN_TEST
    );
    TransactionMapping.createFromDTO(transferTransaction.toJSON());
```

rg911:
Issue: #107 

Fixed a few issues with createFromDTO

### DIFF
--- a/src/infrastructure/transaction/CreateTransactionFromDTO.ts
+++ b/src/infrastructure/transaction/CreateTransactionFromDTO.ts
@@ -30,7 +30,6 @@ import {AggregateTransaction} from '../../model/transaction/AggregateTransaction
 import {AggregateTransactionCosignature} from '../../model/transaction/AggregateTransactionCosignature';
 import {AggregateTransactionInfo} from '../../model/transaction/AggregateTransactionInfo';
 import {Deadline} from '../../model/transaction/Deadline';
-import { LinkAction } from '../../model/transaction/LinkAction';
 import {LockFundsTransaction} from '../../model/transaction/LockFundsTransaction';
 import {ModifyAccountPropertyAddressTransaction} from '../../model/transaction/ModifyAccountPropertyAddressTransaction';
 import {ModifyAccountPropertyEntityTypeTransaction} from '../../model/transaction/ModifyAccountPropertyEntityTypeTransaction';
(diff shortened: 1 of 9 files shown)


src/infrastructure/transaction/CreateTransactionFromDTO.ts
```diff
@@ -118,17 +118,27 @@ export const CreateTransactionFromDTO = (transactionDTO): Transaction => {
  */
 const CreateStandaloneTransactionFromDTO = (transactionDTO, transactionInfo): Transaction => {
     if (transactionDTO.type === TransactionType.TRANSFER) {
+        /**
```
rg911: Not 100% sure if this is the best approach. Basically, if user uses serialisation method `toJson()' it will create a `Message` object rather than encoded message (from the rest endpoint for example). Open for suggestions

src/infrastructure/transaction/CreateTransactionFromDTO.ts
```diff
@@ -349,20 +372,29 @@ const extractTransactionVersion = (version: number): number => {
...
  */
-const extractRecipient = (recipient: string): Address | NamespaceId => {
-    // If bit 0 of byte 0 is not set (like in 0x90), then it is a regular address.
-    // Else (e.g. 0x91) it represents a namespace id which starts at byte 1.
-    const bit0 = convert.hexToUint8(recipient.substr(1, 2))[0];
+const extractRecipient = (recipient: any): Address | NamespaceId => {
+    if (typeof recipient === 'string') {
+        // If bit 0 of byte 0 is not set (like in 0x90), then it is a regular address.
+        // Else (e.g. 0x91) it represents a namespace id which starts at byte 1.
+        const bit0 = convert.hexToUint8(recipient.substr(1, 2))[0];
```
rg911: similar to `PlainMessage` when just toJSON method, it will create either a Address object  or NamespaceId object rather than hex value. Open for suggestion as well

src/infrastructure/transaction/CreateTransactionFromDTO.ts
evias: As proposed on slack, this could be solved in a `extractMessage` auxiliary class.

src/infrastructure/transaction/CreateTransactionFromDTO.ts
```diff
@@ -118,17 +118,27 @@ export const CreateTransactionFromDTO = (transactionDTO): Transaction => {
  */
 const CreateStandaloneTransactionFromDTO = (transactionDTO, transactionInfo): Transaction => {
     if (transactionDTO.type === TransactionType.TRANSFER) {
+        /**
+         * Check if message is encoded (from DTO) or is raw (from JSON)
+         */
+        let message = EmptyMessage;
+        if (transactionDTO.message !== undefined && convert.isHexString(transactionDTO.message.payload)) {
+            message = PlainMessage.createFromDTO(transactionDTO.message.payload);
```
evias: Can you rename to `createFromPayload` ? it is clearer in usage

src/infrastructure/transaction/CreateTransactionFromDTO.ts
```diff
@@ -349,20 +372,29 @@ const extractTransactionVersion = (version: number): number => {
...
-    // read address from encoded hexadecimal notation
-    return Address.createFromEncoded(recipient);
+        // read address from encoded hexadecimal notation
+        return Address.createFromEncoded(recipient);
+    } else if (typeof recipient === 'object') { // Is JSON object
+        if (recipient.hasOwnProperty('address')) {
+            return Address.createFromRawAddress(recipient.address);
+        }
+
+        return new NamespaceId(recipient.id);
```
evias: make sure to check `recipient.id` is available as well no ? (add it in `else if ..`).

There is a throw afterwards but in case `recipient.id` is not present, it will not throw.

src/infrastructure/transaction/CreateTransactionFromDTO.ts
rg911: 👍

rg911: Done!

PR #110, opened 2019-04-08
add EncryptedMessage class

NemTechCN:


src/model/transaction/EncryptedMessage.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Encrypted Message model
+ */
+export class EncryptedMessage extends Message {
+
+    public readonly recipientPublicAccount?: PublicAccount;
+
+    constructor(payload: string,
+                recipientPublicAccount?: PublicAccount){
+        super(1,payload);
```
evias: Please use a constant here `TYPE_ENCRYPTED`.

NemTechCN: ok

PR #120, opened 2019-04-13
Bugfix: handling of unlimited duration mosaics

decentraliser:
Fix: Error when an unlimited duration mosaic MOSAIC_DEFINITION is fetched through `accountHttp.transactions`

Reason: 
`new UInt64_1.UInt64(undefined is returned)`

Bug reproduction:
Use 
`accountHttp.transactions`

on the following account:
`AC1A6E1D8DE5B17D2C6B1293F1CAD3829EEACF38D09311BB3C8E5A880092DE26`

The transaction hash with an unlimited duration mosaic:
`"aggregateHash":"41AA28C4BEC707C09568AAA4C2CF432EE7F8B28E83BC8F5FE24DBF6A92A70042"`

### DIFF
--- a/src/infrastructure/transaction/CreateTransactionFromDTO.ts
+++ b/src/infrastructure/transaction/CreateTransactionFromDTO.ts
@@ -158,7 +158,9 @@ const CreateStandaloneTransactionFromDTO = (transactionDTO, transactionInfo): Tr
             new MosaicProperties(
                 new UInt64(transactionDTO.properties[0].value),
                 (new UInt64(transactionDTO.properties[1].value)).compact(),
-                transactionDTO.properties.length === 3 ? new UInt64(transactionDTO.properties[2].value) : undefined,
+                transactionDTO.properties.length === 3 
+                    ? new UInt64(transactionDTO.properties[2].value) 
+                    : UInt64_1.UInt64.fromUint(0),
             ),
             transactionDTO.signature,
             transactionDTO.signer ? PublicAccount.createFromPublicKey(transactionDTO.signer,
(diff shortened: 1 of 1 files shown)


decentraliser:
This bug is quite blocking, please check ASAP!

decentraliser:
I'm closing this PR as I'm not sure how to produce the expected output but you get the idea! `{ lower: 10, higher: 0 }`

decentraliser:
> None expire mosaic doesn’t need default uint 0 value , you can leave undefined here I believe

In this case shouldn't the `mosaicService.mosaicsAmountViewFromAddress` response updated to `undefined` as well?

rg911:
```
new MosaicProperties(
                new UInt64(transactionDTO.properties[0].value),
                (new UInt64(transactionDTO.properties[1].value)).compact(),
                transactionDTO.properties.length === 3 &&  transactionDTO.properties[2].value ?
                    new UInt64(transactionDTO.properties[2].value) : undefined,
            ),
```

decentraliser:
image
this is my concern, if we return undefined for the transaction service, it will be inconsistent with the mosaic service, but if you are ok with that I can cancel my last commit

rg911:
I would keep the PR and could make the changes I suggested above. We should definitely use `undefined` as per last change request which makes duration optional. I seems to me the rest api is still defaulting to `[0,0]`. I will check 👍

rg911:
@dgarcia360 @Vektrat  
is the rest api using `[0,0]` for default duration on Mosaic definition?

evias:
@Vektrat can you please address latest comment by steven ?

Vektrat:
> @Vektrat can you please address latest comment by steven ?

I ignored it because it was answered and confirmed in slack afaik?

Issue #112, opened 2019-04-09
signWithCosignatories method should not require all cosignatories


rg911:
Aggregated transaction signWithCosignatories current require both initiator Account and cosigners accounts. It is impossible to cosign **aggregated complete transactions** off-chain. 

`signWithCosignatories` method should only take the cosigner's key 

Scenario below

*1. Bob:*

`const aggr = AggregateTransaction.create(...);
const payload = bobAccount.sign(aggr);`

*2. Bob sends the _payload_ to Alice.*

*3. Alice:*

```const aggr = TransactionMapping.createFromPayload(payload)
const signed = aggr.signWithCosignatories([aliceAccount]);
transactionHttp.announce(aggr).subscribe((response) => { console.log("announce: " + signed.hash); }```

rg911:
To recreate the transaction using Bob and Carol cosignatures:

```const aggregateTransaction = AggregateTransaction.createComplete(
    Deadline.create(),
    [
        AtoBTx.toAggregate(accountA.publicAccount),
        BtoATx.toAggregate(accountB.publicAccount),
        CtoATx.toAggregate(accountC.publicAccount)],
    NetworkType.MIJIN_TEST,
    [ new AggregateTransactionCosignature(BobSignature, PublicAccount.createFromPublicKey(Bob's PublicKey),
     new AggregateTransactionCosignature(CarolSignature, PublicAccount.createFromPublicKey(Carol's PublicKey)]
)```

dgarcia360:
Some problems after recreating the transaction using Bob and Carol cosignatures:

* The deadline will differ from the previous transaction created.
* Feels weird to have to define again the same transaction.
* Feels weird to have to sign the transaction again.
* I could not announce the transaction. After running scenario 02.1, line  ``<AggregateTransaction>TransactionMapping.createFromPayload(signedTransactionPayload);``  the console raises ``JavaScript heap out of memory``.

jontey:
Referring to `<AggregateTransaction>TransactionMapping.createFromPayload(signedTransactionPayload);`

Would it make more sense to pass the aggregate transaction rather than the signed transaction?

Actually, if you pass the signed transaction around, then technically it would be possible for the wallet to show how many cosigners have signed the ACT at this point

evias:
[...]

PR #131, opened 2019-04-25
Added #112 sign with cosignatories

rg911:
Issue: #112 

Added a few methods to cope with off chain aggregated complete transaction co signing. 

Please refer to [this scenario](https://github.com/nemtech/nem2-sdk-typescript-javascript/issues/112#issuecomment-481291208) currently been discussed. 

**Problem found from discussion:**
1. Alice needs to create - sign - recreate - sign again
2. `CosignatureTransaction` cannot be created before aggregated transaction's been announced.
3. `signTransactionWithCosignatories(initiatorAccount: Account, cosignatories: Account[])` requires both initiator and the cosigners's account details, which makes it impossible to sign off chain.

**Solution proposed in this PR:**
 01. Alice creates the aggregated tx and serialize it, and generate the hash. Then send to Bob & Carol
      * Use the newly exposed `transaction.serialize()` methods to create transaction payload (without an initial signing). This method will create binary payload with signer and signatures showing as `0000000.....0000`
      * `transaction payload` can then be send to Bob & Carol off the chain.

02. Bob & Carol validate the transaction and cosigns the tx and sends it back to Alice
      * Created a new method `CosignatureTransaction.signTransactionPayload(account, transactionPayload)` which can sign the transaction with its payload. 
      * It will return object `CosignatureSignedTransaction` with signatures Alice expected.

      ```
      const signedTxBob = CosignatureTransaction.signTransactionPayload(accountBob, txPayloadHex);
      const signedTxCarol = CosignatureTransaction.signTransactionPayload(accountCarol, txPayloadHex);
      ```

03. Alice collects the cosignatures, recreate, sign, and announces the transaction
      * Alice will collect those signatures. 

      ``` 
      const cosignatureSignedTransactions = [
            new CosignatureSignedTransaction(signedTxBob.parentHash, signedTxBob.signature, signedTxBob.signer),
            new CosignatureSignedTransaction(signedTxCarol.parentHash, signedTxCarol.signature, signedTxCarol.signer),
        ]; 
      ```

      * Alice then recreate the ACT from payload
      ```
      const recreatedTx = TransactionMapping.createFromPayload(aggregateTransactionPayload) as AggregateTransaction;
      ```

      * To sign the transaction, Alice now use the new method which does not require both initiator Account and cosigners accounts.
      ```
      aggregatedTx.signTransactionGivenSignatures(accountAlice, cosignatureSignedTransactions);
      ```


jontey:
Passing just the hash around is dangerous, as Bob and Carol will not know what they are signing. 

I would suggest that `CosignatureTransaction.signTransactionHashWith` receive a transaction payload rather than just the hash to force developers to pass the entire transaction rather than just the hash (even though the function doesn't do any checks).

The reason behind this is that if the function accepts a hash, most developers will just pass around the hash, making applications that are vulnerable (signing compromised transactions). By forcing the transaction payload, it will notify the developers to think, Hey, maybe I need to get the other side to check if the transaction I'm signing is correct; or make it easier for developers to implement the check (on Bob and Carol's side).

rg911:
> Passing just the hash around is dangerous, as Bob and Carol will not know what they are signing.
> 
> I would suggest that `CosignatureTransaction.signTransactionHashWith` receive a transaction payload rather than just the hash to force developers to pass the entire transaction rather than just the hash (even though the function doesn't do any checks).
> 
> The reason behind this is that if the function accepts a hash, most developers will just pass around the hash, making applications that are vulnerable (signing compromised transactions). By forcing the transaction payload, it will notify the developers to think, Hey, maybe I need to get the other side to check if the transaction I'm signing is correct; or make it easier for developers to implement the check (on Bob and Carol's side).

Fair point, cheers @jontey 👍

evias:
tests failing due to needed library update

evias:
can you rebase @rg911 please ?

rg911:
> can you rebase @rg911 please ?

done

PR #140, opened 2019-05-05
Del encrypted message

NemTechCN:
delet encrypted message,regarding encrypted message. It is an error that the class `EncryptedMessage` got introduced in the SDK using `type=1` because `type` field is actually not supported by catapult-server anymore as can be seen here:https://github.com/nemtech/catapult-server/blob/master/plugins/txes/transfer/src/model/TransferTransaction.h#L50


evias:
I dont think we need this class to be removed.

PR #146, opened 2019-05-10
Task/g138 fix e2e test

rg911:
Issue: #138 

Dependant: Nem2-Lib PR [39](https://github.com/nemtech/nem2-library-js/pull/39)
 
This PR:
- Fixed all e2e tests
- Removed e2e pre-compiled configuration file. Use JSON config `network.conf` instead, so it can be modified at runtime.
- Fixed bug in CreateTranactionFromDto handling `null` message response.
- Fixed HashlockTranasction cannot be created using `Op_Hash_160` algorithm. Padded trailing `0s` to match `SHA3_256` hash length. (See changes in corresponding Nem2-Lib PR)
- Fixed AccountLinkTransaction transaction size too big bug.  (See changes in corresponding Nem2-Lib PR)
- Fixed AccountProperty bugs

On going concerns:
1. For local testing, we have to use bootstrap_server `nemesis_addresses`. Theses addresses may change if local server wiped out
2. `modifyMultisigAccount` Websocket channel does not exists on bootstrap_server. 
3. Cannot test AccountLinkTransaction as it needs harvesting account.
4. Some test will fail if re-run multiple times. E.g. `ModifyMultisigAccountTransaction`


evias:
branch presents build problems due to open nemtech/nem2-library-js#39 PR.

Issue #175, opened 2019-06-16
utf8ToHex in Convert.ts doesn't add insignificant zero while converting 0-15 control characters


yaaccount:
Which will probably render encrypt/decrypt not working on strings containing 0x00-0x0f.
Same issue in convert.js in the js lib.

rg911:
seems working fine for me

`Convert.utf8ToHex('00x0 Hello Word')` => `303078302048656c6c6f20576f7264`

yaaccount:
Try String.fromCodePoint(0x00) + "Hello world"

rg911:
Addressed. well spotted @yaaccount 👍

evias:
fixed in 0.12.3

PR #180, opened 2019-06-17
Issue #175 added contol char handler in convert.ts

rg911:
Fixed bug on convert.utf8ToHex() by adding insignificant zero for control chars

### DIFF
--- a/src/core/format/Convert.ts
+++ b/src/core/format/Convert.ts
@@ -168,12 +168,7 @@ export class Convert {
         const rawString = Convert.rstr2utf8(input);
         let result = '';
         for (let i = 0; i < rawString.length; i++) {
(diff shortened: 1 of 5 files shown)


src/core/format/Convert.ts
```diff
@@ -168,7 +168,12 @@ export class Convert {
         const rawString = Convert.rstr2utf8(input);
         let result = '';
         for (let i = 0; i < rawString.length; i++) {
-            result += rawString.charCodeAt(i).toString(16);
+            if (rawString.charCodeAt(i) < 16) {
```
evias: why not use a `String.padStart()` ?

```
rawString = rawString.padStart(2, '0');
```

rg911: 👍

Issue #189, opened 2019-06-26
Make KeyPair generation and SharedKey derive compatible with NIS1


rg911:
KeyPair generation and shared key derivation for encryption, used Keccak with NIS1 but catapult is using SHA3. 

To make Key generation cross-compatible from the side of the SDK, the SDK needs to allow Keccak algorithm as an option (but default to SHA3)

Catapult on public network will also be using Keccak which is why this change is important to address.

evias:
merged in nemtech/nem2-sdk-typescript-javascript@da9f0fe

PR #190, opened 2019-06-26
[Elephant] Issue #189 Added SignSchema

rg911:
- Introduced the `SignSchema` to make KeyPair generation consistent between Catapult and NIS.
- `SignSchema` contain 2 enum value: `KECCAK_REVERSED_KEY` and `SHA3`. The SDK by default uses `SHA3` which is the Catapult signature schema.

#### Concept

Since Catapult server is using 'SHA3' as default hasher while `NIS1 (current public net)` is `Keccak` based. This change will allow Catapult SDK to be compatible to NIS1's hashes e.g. KeyPair, Address, Encryption, Decryption etc.. 

By default, the SDK uses `SignSchema.SHA3` for hashing unless explicitly specified `KECCAK_REVERSED_KEY`.

- Affected classes:
1. KeyPair
    * `createKeyPairFromPrivateKeyString`
    *  `sign`
    * `verify`
    * `deriveSharedKey`
2. Crypto
    * `encode`
    * `decode`
3. Account
4. Address
    * `createFromPublicKey`
5. PublicAccount
    * `createFromPublicKey`
    * `verifySignature`
6. All transaction signings


src/core/crypto/SignSchema.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ /**
+  * [KECCAK_REVERSED_KEY]: Keccak hash algorithm with key reversed.
```
evias: replace "key reversed" with "reversed private keys"

src/core/crypto/SignSchema.ts
```diff
@@ -0,0 +1,24 @@
...
+  * [SHA3]: Sha3 hash algorithm without key resersed
```
evias: replace text with "SHA3 hash algorithm without key reversal"

src/core/crypto/Utilities.ts
```diff
@@ -201,26 +201,24 @@ export const catapult_crypto = (function() {
...
             const c = nacl;
-            const d = prepareForScalarMult(sk, hashfunc);
+            const d = prepareForScalarMult(sk, hashfunc, signSchema);
 
             // sharedKey = pack(p = d (derived from sk) * q (derived from pk))
             const q = [c.gf(), c.gf(), c.gf(), c.gf()];
             const p = [c.gf(), c.gf(), c.gf(), c.gf()];
             const sharedKey = new Uint8Array(Key_Size);
-            c.unpackneg(q, pk);
+            c.unpack(q, pk);
```
evias: why this change ?

src/core/crypto/nacl_catapult.ts
```diff
@@ -42,22 +42,22 @@ const gf0 = gf(),
     gf1 = gf([1]),
     _121665 = gf([0xdb41, 1]),
     D = gf([0x78a3, 0x1359, 0x4dca, 0x75eb, 0xd8ab, 0x4141, 0x0a4d,
-        0x0070, 0xe898, 0x7779, 0x4079, 0x8cc7, 0xfe73, 0x2b6f, 0x6cee, 0x5203
+        0x0070, 0xe898, 0x7779, 0x4079, 0x8cc7, 0xfe73, 0x2b6f, 0x6cee, 0x5203,
```
evias: linter fixes ?

src/core/crypto/nacl_catapult.ts
```diff
@@ -667,21 +667,21 @@ const unpack25519 = (o, n) => {
...
 
 export const cleanup = (arr) => {
-    for (var i = 0; i < arr.length; i++) {
+    for (let i = 0; i < arr.length; i++) {
         arr[i] = 0;
     }
 };
 
-export const crypto_shared_key_hash = (shared, pk, sk, hashfunc) => {
+export const crypto_shared_key_hash = (shared, pk, sk, hashfunc, hasherType) => {
```
evias: hasherType is actually `signSchema` ?

src/core/crypto/nacl_catapult.ts
```diff
@@ -667,21 +667,21 @@ const unpack25519 = (o, n) => {
...
     const p = [gf(), gf(), gf(), gf()];
 
-    hashfunc(d, sk, 32);
+    hashfunc(d, sk, 32, hasherType);
     d[0] &= 248;
     d[31] &= 127;
     d[31] |= 64;
 
-    let q = [gf(), gf(), gf(), gf()];
+    const q = [gf(), gf(), gf(), gf()];
```
evias: why make it constant ? sure it doesnt break following 3 lines ?

src/core/crypto/nacl_catapult.ts
```diff
@@ -850,3 +850,57 @@ export const unpackneg = (r, p) => {
     M(r[3], r[0], r[1]);
     return 0;
 };
+
+export const unpack = (r, p) => {
```
evias: please explain the need to replace `unpackneg` with this ?

src/model/account/Address.ts
```diff
@@ -25,12 +26,15 @@ export class Address {
...
      * @returns {Address}
      */
-    public static createFromPublicKey(publicKey: string, networkType: NetworkType): Address {
+    public static createFromPublicKey(publicKey: string,
+                                      networkType: NetworkType,
+                                      signSchema = SignSchema.SHA3): Address {
         const publicKeyUint8 = convert.hexToUint8(publicKey);
         const address = AddressLibrary
-            .addressToString(AddressLibrary.publicKeyToAddress(publicKeyUint8, networkType));
+            .addressToString(AddressLibrary.publicKeyToAddress(publicKeyUint8, networkType, signSchema));
```
evias: how does the signature schema affect public key to address algorithm ?

src/model/account/PublicAccount.ts
```diff
@@ -47,13 +47,14 @@ export class PublicAccount {
      * Create a PublicAccount from a public key and network type.
      * @param publicKey Public key
      * @param networkType Network type
+     * @param {SignSchema} signSchema The Sign Schema. (KECCAK_REVERSED_KEY / SHA3)
      * @returns {PublicAccount}
      */
-    static createFromPublicKey(publicKey: string, networkType: NetworkType): PublicAccount {
+    static createFromPublicKey(publicKey: string, networkType: NetworkType, signSchema = SignSchema.SHA3): PublicAccount {
```
evias: see comment about effects on public key derivation / address generation above.

test/core/format/RawAddress.spec.ts
```diff
@@ -145,6 +147,21 @@ describe('address', () => {
             expect(address.isValidAddress(decoded2)).to.equal(true);
             expect(decoded1).to.not.deep.equal(decoded2);
         });
+
+        it('can create address from public key using NIS1 schema', () => {
```
evias: see comments about effects on public key derivations.

src/core/crypto/Utilities.ts
rg911: `unpack` is used NIS1 code. I noticed without using it, the last byte of the hash is different.

src/core/crypto/nacl_catapult.ts
rg911: That's correct, but I don't think we used it anymore, but I will change it anyway, in case it get picked up by the wallet codes.

src/model/account/Address.ts
rg911: hmm, good question. as we need an option to allow choosing either Keccak256 or sha3_256, but not necessarily need to use SignSchema, we can use the existing `HashType` instead. ??

src/model/account/PublicAccount.ts
rg911: see above comments

src/core/crypto/nacl_catapult.ts
rg911: TBH, I haven't figure out why yet. Have been asked core devs on slack. 

Basically, this is only the difference from NIS1 code. We get different hash (last byte) by using 'unpackneg()`, but using `unpack()` works fine. Not quite sure why. The code looks the same to me

src/model/account/Address.ts
rg911: okay, agreed to use `SignSchema` here

src/core/crypto/SignSchema.ts
rg911: done

src/core/crypto/SignSchema.ts
rg911: done

src/core/crypto/nacl_catapult.ts
rg911: done

src/core/crypto/Utilities.ts
evias: what does catapult-server source code say about this?

rg911: where do I find this 😆 ?

src/core/crypto/nacl_catapult.ts
evias: for shared key derivation, `unpack` must be used to be compliant with catapult test vectors.

Issue #166, opened 2019-06-07
[Elephant] Feature: Mosaic Restrictions


evias:
> In catapult, mosaic creators should be allowed to set constraints on accounts that can transact with their defined mosaics.

> This functionality will be enabled by a new mosaic property `Constrainable` that will be disabled by default. Constraints are undesirable for completely fungible tokens (like XEM).

> In order to assign a constraint, a new mosaic constraint transaction is required.

We will be adding a new transaction type `MosaicRestrictionTransaction` to be defined in catapult-server. With following specificities:

- The mosaic creator of the _affected mosaic_ must be signer of this transaction.
- **MosaicRestrictionType** is only applicable to global mosaic constraints (0-Address for Target). When setting on specific account, the constraint type must be `EQ`.

A constraint will be defined by a list of custom properties as defined in following table (counter-check with catapult-server needed for confirmation.)

#### Mosaic Address Restriction
- Version: `1`
- EntityType: `0x4251`

| Field | Type | Description |
| --- | --- | --- | 
| mosaicId | `MosaicId` | Identifier of the mosaic to which the restriction applies |
| restrictionKey | uint64_t | Restriction key |
| targetAdderss | Address `or` 0-Address | Affected address (zero address to set global mosaic instead of account-specific property) |
| previousRestrictionValue | uint64_t | Previous restriction value |
| newRestrictionValue | uint64_t | New restriction value |

#### Mosaic Global Restriction
- Version: `1`
- EntityType: `0x4151`

| Field | Type | Description |
| --- | --- | --- | 
| mosaicId | `MosaicId` | Identifier of the mosaic to which the restriction applies |
| referenceMosaicId | `MosaicId` | Identifier of the mosaic providing the restriction key |
| restrictionKey | uint64_t | Restriction key |
| previousRestrictionValue | uint64_t | Previous restriction value |
| previousRestrictionType | `MosaicRestrictionType` | Previous restriction type |
| newRestrictionValue | uint64_t | New restriction value |
| newRestrictionType | `MosaicRestrictionType` | New restriction type |

#### MosaicRestrictionType
- value: `uint8_t`

| Type | Decimal Value | Description |
| --- | --- | --- | 
| `NONE` | `0` | Uninitialized value indicating no restriction |
| `EQ` | `1` | Must be exact match |
| `NE` | `2` | Must not equal |
| `LT` | `3` | Must be less than |
| `LE` | `4` | Must be less than or equal |
[...]

PR #202, opened 2019-07-12
[Elephant 2] Mosaic Restriction

rg911:
### Mosaic Restriction

- Added `restrictable` in MosaicProperties as a new flag
- Added MosaicGlobalRestriction and MosaicAddressRestriction transactions 
      Including Model,  Schema, Buffer, Builder (Flatbuffer applied)
- Added unit tests and e2e tests for the 2 new txs
- Added serialization handlers (CreateFromDTO / Payload, SerializeToJSON)
- Fixed couple of serialization bugs with MosaicId

#### Note
- MosaicAddressRestrictionTransaction can only be announced with AggregateTransaction
- Empty / Null `ReferenceMosaicId` should use `'0'` in GlobalTransaction
- Empty / Null `PreviousRestrictionValue` should use `FFFFFFFFFFFFFFFF` in AddressRestriction but `0` in GlobalRestriction

#### Schema Reference
https://github.com/nemtech/catbuffer/tree/master/schemas/restriction_mosaic

Issue #166


e2e/infrastructure/TransactionHttp.spec.ts
```diff
@@ -255,7 +255,7 @@ describe('TransactionHttp', () => {
                 Deadline.create(),
                 mosaicId,
                 new MosaicId(UInt64.fromUint(0).toDTO()),
-                UInt64.fromUint(9999),
+                UInt64.fromUint(4444),
```
evias: better to introduce a restriction key enum or is that not possible ?

rg911: > better to introduce a restriction key enum or is that not possible ?

I think the restriction key is free entry. It can be any Uint64 value

evias: ah ok, `magic numbers` logic should be avoided. Also, Why the change from 9999 to 4444 ? 

Better to have the value randomized within range to identify collateral magic number problems easier.

rg911: > ah ok, `magic numbers` logic should be avoided. Also, Why the change from 9999 to 4444 ?
> 
> Better to have the value randomized within range to identify collateral magic number problems easier.
> 
> It's in a unit test so probably not too important but better to change for something that is understood

Yeah, agreed, It's in e2e test . It's trying to match the reference key announced in previous standalone transaction, otherwise this one won't pass. Will update the test value. 👍

e2e/infrastructure/TransactionHttp.spec.ts
```diff
@@ -197,6 +205,118 @@ describe('TransactionHttp', () => {
...
+            return listener.open();
+        });
+        after(() => {
+            return listener.close();
+        });
+        it('aggregate', (done) => {
+            const mosaicAddressRestrictionTransaction = MosaicAddressRestrictionTransaction.create(
+                Deadline.create(),
+                mosaicId,
+                UInt64.fromUint(60641),
```
evias: maybe add a comment to say "why 60641" ? maybe breaking it in its sub part :+1:

rg911: hmm, 60641 has got no meaning. It is just a random number I used here for the test.  Does the value needs to be build from 'something'?

evias: ah ok , I understood that it was a composed value :+1:

Issue #204, opened 2019-07-16
AggregateComplete Transaction cosign with given signatures


rg911:
Bug found in 'signTransactionGivenSignatures()' which always return invalid signature error.

PR #206, opened 2019-07-16
Task/g204 sign transaction given signatures

rg911:
Bug fixed in newly introduced `signTransactionGivenSignatures()`. It is noticed that the signed transaction's signature mismatched with standard cosign methods after applied `generationHash`.

this PR:
- Fixed generationHash should be passed in as byteArray rather than binary in `signTransactionPayload`
- Added e2d test for `signTransactionGivenSignatures()`
- Updated unit tests
- Added signSchema to `signTransactionPayload`
- Fixed typos

### **Process flow changes
It was initially proposed to generate 'plain` (empty signature & signer) payload by the initiator and passed to cosigners. However, the first half of the initiator's signature is required to validate co-signatures. So it is changed that:

1. Alice creates the aggregated transaction and sign it, Then signed payload send to Bob & Carol
2. Bob and Carol cosigns the transaction using `signTransactionPayload` method and sends it back to Alice
3. Alice collects the cosignatures, recreate, sign it using `signTransactionGivenSignatures`, and announces the transaction

Issue : #204


test/model/transaction/AggregateTransaction.spec.ts
```diff
@@ -435,14 +437,20 @@ describe('AggregateTransaction', () => {
...
 
         const signedTransaction = recreatedTx.signTransactionGivenSignatures(accountAlice, cosignatureSignedTransactions, generationHash);
 
         expect(signedTransaction.type).to.be.equal(TransactionType.AGGREGATE_COMPLETE);
         expect(signedTransaction.signer).to.be.equal(accountAlice.publicKey);
         expect(signedTransaction.payload.indexOf(accountBob.publicKey) > -1).to.be.true;
         expect(signedTransaction.payload.indexOf(accountCarol.publicKey) > -1).to.be.true;
+
+        // To make sure that the new cosign method returns the same payload & hash as standard cosigning
+        const standardCosignedTransaction = aggregateTransaction
```
evias: :+1:

PR #239, opened 2019-08-21
Added toString for UInt64 to allow representing a 64-bit integer as a string.

Alexhuszagh:
This is very useful for derived projects from the nem2-typescript-sdk, since it allows representing heights, fees, etc. as strings (useful for user-interfacess, keys for databases, etc.). It supports any radix, and defaults to 10.

### DIFF
--- a/src/model/UInt64.ts
+++ b/src/model/UInt64.ts
@@ -90,18 +90,15 @@ export class UInt64 {
      * @return {string}
      */
     public toString(radix: number = 10): string {
-        var result = '';
-        var high = this.higher;
-        var low = this.lower;
-        while (true) {
-            var mod = (high % radix) * 0x100000000 + low;
+        let result = '';
+        let high = this.higher;
+        let low = this.lower;
+        do {
+            let mod = (high % radix) * 0x100000000 + low;
             high = Math.floor(high / radix);
             low = Math.floor(mod / radix);
             result = (mod % radix).toString(radix) + result;
-            if (!high && !low) {
-                break;
-            }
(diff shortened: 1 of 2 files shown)


src/model/UInt64.ts
```diff
@@ -84,6 +84,27 @@ export class UInt64 {
...
+    /**
+     * Get string representation of number (by default, decimal representation)
+     *
+     * @return {string}
+     */
+    public toString(radix: number = 10): string {
+        var result = '';
+        var high = this.higher;
+        var low = this.lower;
+        while (true) {
```
rg911: while (true)?? Don't see it as good practice

src/model/UInt64.ts
```diff
@@ -84,6 +84,27 @@ export class UInt64 {
         return uint64.toHex(this.toDTO());
     }
 
+    /**
+     * Get string representation of number (by default, decimal representation)
+     *
+     * @return {string}
+     */
+    public toString(radix: number = 10): string {
+        var result = '';
```
rg911: can we change to `const` or `let` instead of var?

Alexhuszagh: Sure, definitely. Let me edit that commit.

src/model/UInt64.ts
Alexhuszagh: I could change it to a `do while` loop: we need at least 1 iteration, so the break condition is at the end.

Alexhuszagh:
@rg911, Are the submitted changes better?

PR #248, opened 2019-09-10
fix #247

### LINKED ISSUE #247: mosaicsAmountViewFromAddress calling mosaicHttp.getMosaics once / mosaic
Here `mosaicsAmountView `sends mosaicIds one by one to `mosaicsView`
Then `mosaicsView` makes one `mosaicHttp.getMosaics` call per mosaics
This results in calling the endpoints `/mosaic` and `/network` one time per mosaic and the client eventually hangs when there are too many `mosaicId`

All `mosaicId` should be passed to `getMosaics` in an array

https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/c8259f9590ac9fa7d954349b10fe55a71ff52911/src/service/MosaicService.ts#L67

decentraliser:
This does the trick for me in the compiled file, it's much faster
I didn't try to build it on the SDK

working compiled version:

    mosaicsAmountView(mosaics) {
        return rxjs_1.from(mosaics)
            .pipe(
                operators_1.toArray(),
                operators_1.concatMap(mosaics =>
                    rxjs_1.zip(
                        rxjs_1.of(mosaics).pipe(operators_1.flatMap(x=>x)),
                        rxjs_1.of(mosaics)
                            .pipe(
                                operators_1.flatMap(x => x),
                                operators_1.map(mosaic => new MosaicId_1.MosaicId(mosaic.id.id.toDTO())),
                                operators_1.toArray(),
                                operators_1.mergeMap((mosaic) => this.mosaicsView(mosaic)),
                                operators_1.flatMap(x=>x)
                            )
                    ).pipe(
                        operators_1.map(([mosaic, mosaicView]) => {
                            return new MosaicAmountView_1.MosaicAmountView(mosaicView.mosaicInfo, mosaic.amount)
                           })
                        ),
                ),
                operators_1.toArray(),
            )
    }

### DIFF
--- a/src/service/MosaicService.ts
+++ b/src/service/MosaicService.ts
@@ -62,16 +62,30 @@ export class MosaicService {
      * @returns {Observable<MosaicAmountView[]>}
      */
     mosaicsAmountView(mosaics: Mosaic[]): Observable<MosaicAmountView[]> {
-        return observableOf(mosaics).pipe(
-            mergeMap((_) => _),
-            mergeMap((mosaic: Mosaic) => this.mosaicsView([new MosaicId(mosaic.id.id.toDTO())]).pipe(
-                filter((_) => _.length !== 0),
-                map<MosaicView[], MosaicAmountView>((mosaicViews) => {
-                    return new MosaicAmountView(mosaicViews[0].mosaicInfo, mosaic.amount);
-                }),
-            toArray())));
+        return from(mosaics)
+            .pipe(
+                toArray(),
+                concatMap(mosaics =>
+                    zip(
+                        of(mosaics).pipe(flatMap((_) => _)),
+                        of(mosaics)
+                            .pipe(
+                                flatMap((_) => _),
+                                map(mosaic => new MosaicId(mosaic.id.id.toDTO())),
+                                toArray(),
+                                mergeMap((mosaic) => this.mosaicsView(mosaic)),
+                                flatMap((_) => _),
+                            )
+                    ).pipe(
+                        map(([mosaic, mosaicView]) => {
+                            return new MosaicAmountView(mosaicView.mosaicInfo, mosaic.amount)
+                        })
+                    ),
+                ),
+                toArray(),
+            )
     }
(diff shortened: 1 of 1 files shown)


src/service/MosaicService.ts
```diff
@@ -62,16 +62,30 @@ export class MosaicService {
      * @returns {Observable<MosaicAmountView[]>}
      */
     mosaicsAmountView(mosaics: Mosaic[]): Observable<MosaicAmountView[]> {
-        return observableOf(mosaics).pipe(
-            mergeMap((_) => _),
-            mergeMap((mosaic: Mosaic) => this.mosaicsView([new MosaicId(mosaic.id.id.toDTO())]).pipe(
```
evias: this in fact will call the networkType observable for any mosaics. What about fixing the issue in this method `mosaicsView` (fix: get network type only once for whole Http service with said endpoint.) https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/1e4814a1d47b6127ac5c5489c9f9ce985cb50c37/src/service/MosaicService.ts#L51

Seems like the fix is very cumbersome and could be refactored using mergeMap or other operators.

decentraliser:
https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/1e4814a1d47b6127ac5c5489c9f9ce985cb50c37/src/service/MosaicService.ts#L76

The mosaic view looks fine to me, if you pass an array to it, it will just
call the network type once

src/service/MosaicService.ts
decentraliser: The mosaic view looks fine to me, if you pass an array to it, it will just
call the network type once:
https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/1e4814a1d47b6127ac5c5489c9f9ce985cb50c37/src/service/MosaicService.ts#L76

Yes, my proposition is not elegant... and I actually just noticed that it does not work, because getMosaics() does not return the mosaics in the same order.

Issue #253, opened 2019-09-17
[Elephant3] Signature schema refactoring


rg911:
Based on the latest confirmation from the Core devs. We should be able to use NetworkType to identify signature schema. 

- Public(Test) => Keccak
- Mijin(Test) => SHA3
- Reverse privateKey is not required anymore.

PR #254, opened 2019-09-18
JAV-18 [Github #253] Signature Schema Refactoring

rg911:
#### Signature Schema Refactoring

- Signature schema (NIS1 vs Catapult) can be identified by `NetworkType` (based on confirmation from Core on [Sig-API]. **Signatures of methods which previously used `SignSchema` have been refactored to use `NetworkType` instead**.

      - NIS1: Public(Test) => Keccak (No key reversal any more)
      - Catapult: MIjin(Test) => SHA3

- KeyPair generation under NIS1 does **NOT** reverse the private key bytes anymore.  Keys & Address generated / derived under NIS1 schema in the SDK will not be compatible with current public net any more. 
- Fixed all unit tests (due to key reversal removed)


src/core/crypto/Crypto.ts
```diff
@@ -219,16 +221,16 @@ export class Crypto {
      * @param {string} msg - A text message
      * @param {Uint8Array} iv - An initialization vector
      * @param {Uint8Array} salt - A salt
-     * @param {SignSchema} signSchema The Sign Schema. (KECCAK_REVERSED_KEY / SHA3)
+     * @param {NetworkType} networkType - Catapult network identifier
```
fboucquez: I took a slightly different approach.  The crypto component/module knows about SignSchema not about network type. The layer converting from NetworkType to SignSchema is the model layer (Transaction, Addresses, Account, KeyPair, etc) It's more a philosophical comment.

src/core/crypto/KeyPair.ts
```diff
@@ -74,21 +66,16 @@ export class KeyPair {
...
         if (Utility.Key_Size !== publicKey.length) {
             throw Error(`public key has unexpected size: ${salt.length}`);
         }
-        let secretKey = keyPair.privateKey;
-        // KECCAK_REVERSED_KEY uses reversed private key.
-        if (signSchema === SignSchema.KECCAK_REVERSED_KEY) {
-            secretKey = convert.hexToUint8Reverse(convert.uint8ToHex(secretKey));
-        }
-        return Utility.catapult_crypto.deriveSharedKey(salt, secretKey, publicKey, Utility.catapult_hash.func, signSchema);
+        return Utility.catapult_crypto.deriveSharedKey(salt, keyPair.privateKey, publicKey, Utility.catapult_hash.func, networkType);
```
fboucquez: I would keep the reverse hooks, maybe create a utility method that reverse if necessary (called every time there was a hexToUint8Reverse). Just in case you want to add the reverse stuff in the future.

src/infrastructure/transaction/CreateTransactionFromPayload.ts
```diff
@@ -46,13 +46,11 @@ import { TransactionBuilder } from '../catbuffer/TransactionBuilder';
...
  * @param payload - The transaction binary data
  * @param isEmbedded - Is the transaction an embedded inner transaction
- * @param signSchema - The Sign Schema. (KECCAK_REVERSED_KEY / SHA3)
  * @returns {Transaction | InnerTransaction}
  * @constructor
  */
 export const CreateTransactionFromPayload = (payload: string,
-                                             isEmbedded = false,
-                                             signSchema = SignSchema.SHA3): Transaction | InnerTransaction => {
+                                             isEmbedded = false): Transaction | InnerTransaction => {
```
fboucquez: Much simplier now that we know sign schema is derived from network type

test/core/format/RawAddress.spec.ts
```diff
@@ -369,11 +325,11 @@ describe('address', () => {
             ];
 
             const Addresses = [
-                'NDIPRQMB3HT7A6ZKV7HOHJQM7JHX6H3FN7DBVFHB',
-                'NC65QJI4OWTUFJNQ2IDVOMUTE7IDI2EGEFP5QME7',
-                'NCBC4VAQBVSB4J5J2PTFM7OUY5CYDL33VW7RKW7K',
-                'NBLW3CQPBGPCFAXG4XM5GDEVLPESCPDPFO7NUCOM',
-                'NA5RDU36TKBTW4KVSSPD7PT5YTUMD7OIJFMMIEEQ',
+                'MDIPRQMB3HT7A6ZKV7HOHJQM7JHX6H3FN5YHHZMD',
```
fboucquez: Maybe you can load the test cases directly from the vector files. I've added that to Java sdk.

src/core/crypto/Crypto.ts
rg911: Had the same thought as well. First draft didn't touch the signSchema on the core component level. But also consider that signSchema belongs to `NetworkSetting` as Jag mentioned in sig-api. So finally decided to use NetworkType. I think your suggestion makes sense as well. I will change it back.

src/core/crypto/KeyPair.ts
rg911: Not sure about this. Let's confirm with core first. If the decision is non-reversal I don't think we should keep it. 

The 'hexToUint8Reverse' definitely stays, as will probably need it at some point. E.g. the UInt64 is using `Little Endian`.

test/core/format/RawAddress.spec.ts
rg911: Good point, but the size of the test vector json files is a concern. To keep all unit tests run offline, need to copy the files to the project.

src/core/format/RawAddress.ts
```diff
@@ -105,10 +106,11 @@ export class RawAddress {
     /**
      * Determines the validity of a decoded address.
      * @param {Uint8Array} decoded The decoded address.
-     * @param {SignSchema} signSchema The Sign Schema. (KECCAK_REVERSED_KEY / SHA3)
+     * @param {number} networkIdentifier The network identifier.
```
evias: what is network identifier ? Is it a `networkType` ? or maybe do we need an added model NetworkIdentifier ?

rg911: I think's it's a typo. should read NetworkType

PR #257, opened 2019-09-20
V0.13.2 release 

rg911:


dgarcia360:
Thanks for completing the changelog 🥇 The PR looks correct, but please coordinate with the REST team to also tag a partial e3 compatible version on their side so that the development of other components can move forward.

Issue #271, opened 2019-09-25
Reapply OpenAPI generated codes


rg911:
OpenAPI updated with bugs fixed. Need to regenerate the codes apply to SDK

Issue #266, opened 2019-09-23
Add metadata endpoint http repository


rg911:
New metadata endpoint added in rest which need to be implemented.

PR #272, opened 2019-09-26
JAV-42 & JAV-45 Metadata and MosaicRestriction endpoints implementation

rg911:
- Metadata endpoints implemented.
- MosaicRestriction endpoints implemented.
-------
- Latest OpenApi generated code applied (Without MosaicRestriction)
- Added unit  and e2e tests
- Fixed `Namespace/names` endpoint bug (openApi doc issue reported)

Issue: #266 #271


src/infrastructure/MetadataHttp.ts
```diff
@@ -0,0 +1,267 @@
...
+        return new Metadata(
+            metadata.id,
+            new MetadataEntry(
+                metadataEntry.compositeHash,
+                metadataEntry.senderPublicKey,
+                metadataEntry.targetPublicKey,
+                UInt64.fromHex(metadataEntry.scopedMetadataKey),
+                metadataEntry.metadataType.valueOf(),
+                metadataEntry.valueSize,
+                metadataEntry.value,
```
fboucquez: When a metadata transaction is announced, the user uses plain text. The sdk hides all the seralization details. That's great! 
But when getting the metadata entry from the endpoint, the value is in hex. The user is charged for converting the hex back to the provided value. The value hex -> plain text conversion should be done here and not in the client like the e2e tests.

e2e/infrastructure/MetadataHttp.spec.ts
```diff
@@ -0,0 +1,411 @@
...
+    describe('getAccountMetadata', () => {
+        it('should return metadata given a NEM Address', (done) => {
+            metadataHttp.getAccountMetadata(accountAddress)
+                .subscribe((metadata) => {
+                    expect(metadata.length).to.be.greaterThan(0);
+                    expect(metadata[0].metadataEntry.scopedMetadataKey.toString()).to.be.equal('5');
+                    expect(metadata[0].metadataEntry.senderPublicKey).to.be.equal(account.publicKey);
+                    expect(metadata[0].metadataEntry.targetPublicKey).to.be.equal(account.publicKey);
+                    expect(metadata[0].metadataEntry.targetId).to.be.undefined;
+                    expect(Convert.decodeHex(metadata[0].metadataEntry.value)).to.be.equal('Test account meta value');
```
fboucquez: Client should not convert the value. The value should be already in plain text.

test/core/utils/TransactionMapping.spec.ts
```diff
@@ -1122,6 +1122,6 @@ describe('TransactionMapping - createFromDTO (Transaction.toJSON() feed)', () =>
         expect(transaction.scopedMetadataKey.toString()).to.be.equal(UInt64.fromUint(1000).toString());
         expect(transaction.valueSizeDelta).to.be.equal(1);
         expect(transaction.targetNamespaceId.toHex()).to.be.equal(new NamespaceId([2262289484, 3405110546]).toHex());
-        expect(Convert.uint8ToHex(transaction.value)).to.be.equal(Convert.uint8ToHex(new Uint8Array(10)));
+        expect(transaction.value).to.be.equal('Test Value');
```
fboucquez: The value in transaction is already plain text :)

src/infrastructure/model/resolutionEntryDTO.ts
```diff
@@ -42,7 +42,7 @@ export class ResolutionEntryDTO {
         {
             "name": "resolved",
             "baseName": "resolved",
-            "type": "AnyOfstringstring"
+            "type": "any"
```
fboucquez: instead of patching the generated code, I'm patching the open api spec. The generated code in java is untouched, not even committed. Maybe is better?

rg911: Will probably refactor it after RC

PR #288, opened 2019-10-07
Implement KeyGenerator

contractormario:
PR adds `core/format/KeyGenerator`. Usage:

`const u64 = KeyGenerator.fromString('key_name') // returns UInt64`

### DIFF
--- a/src/core/format/KeyGenerator.ts
+++ b/src/core/format/KeyGenerator.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 NEM
+ * Copyright 2019 NEM
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,18 +19,16 @@ import { sha3_256 } from 'js-sha3';
 
 export class KeyGenerator {
     /**
+     * Generate UInt64 from a string
+     * @param {string} input Input string
      * @returns {UInt64} Deterministic uint64 value for the given string
      */
-    public static fromString(input: string) {
+    public static fromString(input: string): UInt64 {
         if (input.length === 0) {
             throw Error(`Input must not be empty`);
         }
         if (input.length > 1024) {
             throw Error(`Input exceeds 1024 characters (has ${input.length})`);
-        }
-        const format = /^[a-zA-Z0-9_]+$/gm;
-        if (input.match(format) === null) {
-            throw Error(`Input has invalid format (accepted characters: a-z, A-Z, 0-9, _)`);
         }
         const hex = sha3_256(input)
         return UInt64.fromHex(hex.substr(0, 16))
--- a/src/core/format/KeyGenerator.ts
+++ b/src/core/format/KeyGenerator.ts
@@ -23,14 +23,12 @@ export class KeyGenerator {
      * @param {string} input Input string
      * @returns {UInt64} Deterministic uint64 value for the given string
      */
-    public static fromString(input: string): UInt64 {
+    public static generateUInt64Key(input: string): UInt64 {
         if (input.length === 0) {
             throw Error(`Input must not be empty`);
         }
-        if (input.length > 1024) {
-            throw Error(`Input exceeds 1024 characters (has ${input.length})`);
-        }
-        const hex = sha3_256(input)
-        return UInt64.fromHex(hex.substr(0, 16))
+        const buf = sha3_256.arrayBuffer(input);
+        const result = new Uint32Array(buf);
+        return new UInt64([result[0], (result[1] | 0x80000000) >>> 0]);
     }
 }
--- a/src/core/format/KeyGenerator.ts
+++ b/src/core/format/KeyGenerator.ts
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2018 NEM
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { UInt64 } from '../../model/UInt64';
+import { sha3_256 } from 'js-sha3';
+
+export class KeyGenerator {
+    /**
+     * @returns {UInt64} Deterministic uint64 value for the given string
```
rg911: doc is wrong. missing description and and params

src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public static fromString(input: string) {
+        if (input.length === 0) {
+            throw Error(`Input must not be empty`);
+        }
+        if (input.length > 1024) {
```
rg911: not all keys have the same restriction. the check should be case specific

src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,38 @@
...
+            throw Error(`Input exceeds 1024 characters (has ${input.length})`);
+        }
+        const format = /^[a-zA-Z0-9_]+$/gm;
+        if (input.match(format) === null) {
+            throw Error(`Input has invalid format (accepted characters: a-z, A-Z, 0-9, _)`);
+        }
+        const hex = sha3_256(input)
```
rg911: why hash the input string? should only need to buffer the string

src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,38 @@
...
+        return UInt64.fromHex(hex.substr(0, 16))
```
rg911: should return hex string as per recent rest changes.

src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+import { UInt64 } from '../../model/UInt64';
+import { sha3_256 } from 'js-sha3';
+
+export class KeyGenerator {
+    /**
+     * @returns {UInt64} Deterministic uint64 value for the given string
+     */
+    public static fromString(input: string) {
```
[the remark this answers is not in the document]
fboucquez: Missing return type in the signature

src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (input.length === 0) {
+            throw Error(`Input must not be empty`);
+        }
+        if (input.length > 1024) {
+            throw Error(`Input exceeds 1024 characters (has ${input.length})`);
+        }
+        const format = /^[a-zA-Z0-9_]+$/gm;
```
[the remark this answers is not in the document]
contractormario: Removed

src/core/format/KeyGenerator.ts
contractormario: Fixed

src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2018 NEM
```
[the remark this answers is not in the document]
contractormario: Fixed

src/core/format/KeyGenerator.ts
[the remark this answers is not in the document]
contractormario: What name do you suggest, generateUint64? generateKey?

src/core/format/KeyGenerator.ts
contractormario: Not sure what you mean?

src/core/format/KeyGenerator.ts
contractormario: longer strings with identical first 16 characters would produce the same u64. hashing guarantees uniqueness

src/core/format/KeyGenerator.ts
rg911: If I remember correctly only the `metadata.value` require this check. Don't think we need this check at all. do we?

src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public static fromString(input: string): UInt64 {
+        if (input.length === 0) {
+            throw Error(`Input must not be empty`);
+        }
+        if (input.length > 1024) {
+            throw Error(`Input exceeds 1024 characters (has ${input.length})`);
+        }
+        const hex = sha3_256(input)
+        return UInt64.fromHex(hex.substr(0, 16))
```
rg911: if you check the definition of UInt64, it uses 2 Uint32 (low / high). I would convert the hash buffer array to uint32 array and return the new Uint64( ['low', 'high']). Not quite sure about the substring here.

I would do something like
```
const result = new Uint32Array(hex.arrayBuffer());
return new UInt64([result[0], (result[1] | 0x80000000) >>> 0]);
```

src/core/format/KeyGenerator.ts
```diff
@@ -0,0 +1,36 @@
...
+import { UInt64 } from '../../model/UInt64';
+import { sha3_256 } from 'js-sha3';
+
+export class KeyGenerator {
+    /**
+     * Generate UInt64 from a string
+     * @param {string} input Input string
+     * @returns {UInt64} Deterministic uint64 value for the given string
+     */
+    public static fromString(input: string): UInt64 {
```
rg911: Shall we call it `generateUInt64Key` instead? 
@evias @dgarcia360

Issue #295, opened 2019-10-09
Metadata and Mosaic Restriction transactions do not support aliased mosaics


dgarcia360:
The serialization schemas define that the next fields can be aliased.

* https://github.com/nemtech/catbuffer/blob/master/schemas/metadata/mosaic_metadata.cats#L12
* https://github.com/nemtech/catbuffer/blob/master/schemas/restriction_mosaic/mosaic_address_restriction.cats#L6
* https://github.com/nemtech/catbuffer/blob/master/schemas/restriction_mosaic/mosaic_global_restriction.cats#L7-10

The NEM2-SDK should accept ``NamespaceId`` for these fields as well as the current ``MosaicId`` type.

rg911:
@Vektrat @martinayora 
Does rest support query namespaceId on above endpoints?

Vektrat:
> @Vektrat @martinayora
> Does rest support query namespaceId on above endpoints?

what endpoints? david is talking about transactions here, i believe we are talking about announcing transactions?

rg911:
> > @Vektrat @martinayora
> > Does rest support query namespaceId on above endpoints?
> 
> what endpoints? david is talking about transactions here, i believe we are talking about announcing transactions?

get mosaicRestriction for example. should it be allow to use namespaceId?

Vektrat:
i'm not sure, depends on whether the mongodb collection is a mosaicId as stated, or if it can be both, which i doubt - but i still think david is talking about transactions, not endpoints

dgarcia360:
Yes, I'm referring to transactions at the moment. 

The endpoints contract does not have to change, as you can look up for the resolution with another endpoint. However, we might indicate in the open-api spec if the state stores the mosaic resolved (MosaicId) or not (NamespaceId). We can confirm the latest once we try to send a transaction using an aliased mosaic.

Vektrat:
since both mosaicId and namespaceId are uint64, obviously both fields are supported and forwarded to core **in transactions**

PR #297, opened 2019-10-11
JAV-64 Unresolved Mosaic and Address in transactions

rg911:
- Created UnresolvedMapping Util class
- Updated Transactions signature to allow NamespaceId (MosaicRestrictions, MosaicMetadata)
- Update serialization (DTO & Binary payload) to cope with alias
- Created / Added unit and integration tests for above changes. 

Issue: #295


src/model/transaction/MosaicAddressRestrictionTransaction.ts
```diff
@@ -173,6 +175,22 @@ export class MosaicAddressRestrictionTransaction extends Transaction {
                byteTargetAddress + bytePreviousRestrictionValue + byteNewRestrictionValue;
     }
 
+    /**
+     * Return the string notation for the set recipient
+     * @internal
+     * @returns {string}
+     */
+    public targetAddressToString(): string {
```
evias: why name it `target` instead of recipient?

rg911: this is to match the wg-standard and the catabuffer schema definition

PR #304, opened 2019-10-12
JAV-38  Inner transaction singer delegation

rg911:
- Changed signer param to be optional on `toAggregate` when adding inner transactions. If not signer in inner transaction not provided, the main Aggregate transaction signer will be delegated.
- This change **ONLY** applies on **AggregateComplete** transaction
- Added validation if no signer (inner tx) provided in AggregateBond transaction
- Added tests


rg911:
@dgarcia360 @evias

PR #327, opened 2019-10-29
TS-77 filtering listener's status message by subscribed address.

fboucquez:
Hi @rg911 

I've added the filter by address to the listener's status subscription, 

https://github.com/nemtech/catapult-rest/pull/174 fix has been merged. Remember to get the latest catapult-rest when working as the new status listener requires the latest rest.

### DIFF
--- a/src/infrastructure/Listener.ts
+++ b/src/infrastructure/Listener.ts
@@ -127,7 +127,7 @@ export class Listener {
     /**
      * @internal
      *
-     * This this method handles one incoming message from the web socket and it dispatch it to the right message subject listener.
+     * This method handles one incoming message from the web socket and it dispatches it to the message subject listener.
      *
      * @param message the object payload.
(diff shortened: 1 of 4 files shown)


src/infrastructure/Listener.ts
```diff
@@ -114,58 +114,65 @@ export class Listener {
...
-                    }
+                    this.handleMessage(message, resolve);
                 };
             } else {
                 resolve();
             }
         });
     }
 
+    handleMessage(message: any, resolve) {
```
rg911: Doc, make it `Private` and `@Internal`

fboucquez: can it be called from the unit test if it's private?

rg911: ah. okay, you added unit tests. so just need to make it `internal` then I think

PR #330, opened 2019-10-29
TS-81 Address.isValidAddress methods

fboucquez:
Hi @rg911 

I've revisited the is valid address methods. The solution is that RawAddress class handles the low-level byte array validation doing the hashes and checksum checks. The Address class handles the high-level validation, with 2 methods isValidPlainAddress and isValidEncodedAddress that receive strings. The address delegates the validation to RawAddress. I've also added Address.encoded() so you can get the hex string version of the address. 

Have a look at let me know what do you think.

### DIFF
--- a/src/model/account/Address.ts
+++ b/src/model/account/Address.ts
@@ -13,8 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Convert, RawAddress } from '../../core/format';
-import {NetworkType} from '../blockchain/NetworkType';
+import { Convert, RawAddress, RawArray } from '../../core/format';
+import { NetworkType } from '../blockchain/NetworkType';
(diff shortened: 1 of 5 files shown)


src/model/account/Address.ts
```diff
@@ -65,12 +67,41 @@ export class Address {
...
+    }
+
+
+    /**
+     * Determines the validity of an raw address string.
+     * @param {string} rawAddress The raw address string. Expected format SCHCZBZ6QVJAHGJTKYVPW5FBSO2IXXJQBPV5XE6P
+     * @param {NetworkType} networkType The network identifier.
+     * @returns {boolean} true if the raw address string is valid, false otherwise.
+     */
+    public static isValidRawAAddress = (rawAddress: string, networkType: NetworkType): boolean => {
```
rg911: Typo here on the method name

fboucquez: Fixed

Issue #348, opened 2019-11-12
MosaicId can be instantiated with undefined


decentraliser:
https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/dc6135d141b4d3de75fac8f64ed38705c92d59ce/src/model/mosaic/MosaicId.ts#L52

`new MosaicId(undefined)` does not throw an error. As a consequence, it is possible to create a transaction with an empty Mosaic object.

![image](https://user-images.githubusercontent.com/36499647/68645379-d065a180-0552-11ea-870f-4ad1d2d8ac2e.png)
![image](https://user-images.githubusercontent.com/36499647/68645400-e70bf880-0552-11ea-8ecd-13820233cd36.png)

Issue #319, opened 2019-10-23
NetworkHttp should use Network Routes and not Node Routes


fboucquez:
This can be possible once all the possible string enum values have been defined.
nemtech/nem2-openapi#44

dgarcia360:
@fboucquez I guess we could close this issue due to the roll-back.

PR #354, opened 2019-11-15
Added GetNetworkName and other small fixes

rg911:
- Added `getNetworkName()` in `NetworkHttp` which return the name and description of current network. Kept the `getNetworkType()` which return the current network identifier from `node/info`
- Added validation in `MosaicId` construction for undefined id (though the undefined value should not get passed via type check by Typescript)
- Fixed small bugs in serializeToJson (MosaicDefinitionTx)

Issue: #348, #319


src/infrastructure/NetworkHttp.ts
```diff
@@ -58,4 +63,18 @@ export class NetworkHttp extends Http implements NetworkRepository {
...
+     * Get current network type name and description
+     *
+     * @return current network type name and description
+     */
+    public getNetworkName(): Observable<NetworkName> {
+        return observableFrom(this.networkRouteApi.getNetworkType()).pipe(
+            map((response: { response: ClientResponse; body: NetworkTypeDTO; }) => {
+                return new NetworkName(response.body.name, response.body.description);
+            }),
+            catchError((error) =>  throwError(this.errorHandling(error))),
```
fboucquez: Remove the inner response. Use the `body.name` and `body.description` directly. It's type-safe. You don't need the `: NetworkTypeDto`, ts automatically knows that the body is typed to that NetworkTypeDto. 

You can also remove curly braces and the return world

I have the same method in my PR. I called NetworkInfo with a name and description. 

https://github.com/nemtech/nem2-sdk-typescript-javascript/pull/342/commits/3637bff3f7a66db7d3c169a34c417b1fb4920880#diff-5cb1ce66c671496f743ab2481416cf08R73

src/infrastructure/transaction/SerializeTransactionToJSON.ts
```diff
@@ -130,7 +130,7 @@ export const SerializeTransactionToJSON = (transaction: Transaction): any => {
         case TransactionType.MOSAIC_DEFINITION:
             return {
                 nonce: (transaction as MosaicDefinitionTransaction).nonce,
-                mosaicId: (transaction as MosaicDefinitionTransaction).mosaicId.toHex(),
+                id: (transaction as MosaicDefinitionTransaction).mosaicId.toHex(),
                 flags: (transaction as MosaicDefinitionTransaction).flags.getValue(),
                 divisibility: (transaction as MosaicDefinitionTransaction).divisibility,
```
fboucquez: Why the multiple downcasting? Should it be done once?

src/model/mosaic/MosaicId.ts
```diff
@@ -50,6 +50,9 @@ export class MosaicId {
      * @param id
      */
     constructor(id: string | number[]) {
+        if (id === undefined) {
```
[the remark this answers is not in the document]
rg911: not possible IMO.  but since it got reported I just added it for extra check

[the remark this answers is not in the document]
evias: validation by type seems to be problematic due to different compilation modes with TypeScript ...

OK for this PR, **note that this probably means that we need to fix _a lot of different methods to validate parameter input_**.

src/infrastructure/NetworkHttp.ts
rg911: will apply on rest HTTP clients in other PR

PR #356, opened 2019-11-19
refactor Transaction.createTransactionHash and add unit test for netw…

evias:
### Changed

- Refactor of Transaction.createTransactionHash for clarity and use protocol identical transaction creation flow as seen in [TransactionExtensions](https://github.com/nemtech/catapult-server/blob/master/sdk/src/extensions/TransactionExtensions.cpp#L46)

### DIFF
--- a/src/model/transaction/Transaction.ts
+++ b/src/model/transaction/Transaction.ts
@@ -36,9 +36,32 @@ export abstract class Transaction {
     /**
      * Transaction header size
      *
+     * Included fields are `size`, `verifiableEntityHeader_Reserved1`,
+     * `signature`, `signerPublicKey` and `entityBody_Reserved1`.
+     *
      * @var {number}
      */
-    public static readonly Header_Size = 8 + 64 + 32 + 4;
+    public static readonly Header_Size: number = 8 + 64 + 32 + 4;
+
+    /**
+     * Index of the transaction *type*
+     *
+     * Included fields are the transaction header, `version`
+     * and `network`
+     *
+     * @var {number}
+     */
+    public static readonly Type_Index: number = Transaction.Header_Size + 2;
+
+    /**
+     * Index of the transaction *body*
+     *
+     * Included fields are the transaction header, `version`,
+     * `network`, `type`, `maxFee` and `deadline`
+     *
+     * @var {number}
+     */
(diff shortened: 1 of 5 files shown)


src/model/transaction/Transaction.ts
```diff
@@ -81,29 +88,54 @@ export abstract class Transaction {
...
+        const publicKey: Uint8Array = transactionBytes.slice(8 + 64, 8 + 64 + 32);
+
+        // 3) add generationHash (32 bytes)
+        const generationHashIdx: number = pubKeyIdx + publicKey.length;
+        const generationHash: Uint8Array = Uint8Array.from(generationHashBuffer);
+
+        // 4) add transaction data without header (EntityDataBuffer)
+        // @link https://github.com/nemtech/catapult-server/blob/master/src/catapult/model/EntityHasher.cpp#L30
+        const transactionBodyIdx: number = generationHashIdx + generationHash.length;
+        const transactionBody: Uint8Array = transactionBytes.slice(Transaction.Header_Size);
```
rg911: Not sure this will work on the AggregateTransaction specifically. As the new `signingBytes` for Aggregated changed to only include the `merkleTransactionHash` field but not rest of the payload

evias: changed will be applied with next push

evias:
on hold because of Keccak usage in transaction hash, maybe not necessary will update PR accordingly

src/model/transaction/Transaction.ts
```diff
@@ -81,29 +111,68 @@ export abstract class Transaction {
...
+          + transactionBody.length,
+        );
+        entityHashBytes.set(signatureR, 0);
+        entityHashBytes.set(publicKey, pubKeyIdx);
+        entityHashBytes.set(generationHash, generationHashIdx);
+        entityHashBytes.set(transactionBody, transactionBodyIdx);
 
-        return Convert.uint8ToHex(hash);
+        // 6) create SHA3 or Keccak hash depending on `signSchema`
+        const signSchema: SignSchema = SHA3Hasher.resolveSignSchema(networkType);
```
rg911: As gimre confirmed with me on slack, TransactionHash is always using sha3_256, I think we don't need resolving the signSchema anymore

PR #358, opened 2019-11-20
Apply OpenAPI client code changes for F2

rg911:
- Latest OpenAPI spec for F2 applied
- Fixed Receipt serialization bug 
- Added missing e2e tests for MosaicOwner endpoints


src/infrastructure/MetadataHttp.ts
```diff
@@ -259,7 +231,7 @@ export class MetadataHttp extends Http implements MetadataRepository {
                 metadataEntry.targetPublicKey,
                 UInt64.fromHex(metadataEntry.scopedMetadataKey),
                 metadataEntry.metadataType.valueOf(),
-                metadataEntry.valueSize,
+                Convert.decodeHex(metadataEntry.value).length,
```
fboucquez: Should we just remove the size from the transaction model? I took it out in Java

evias: what is the reasoning behind removing it ?

rg911: this size is not defined in OpenAPI, and I believe because it is not returning from the rest.

rg911: It can be easily calculated from `value.length` I think. as how the SDK is doing here

PR #359, opened 2019-11-21
Refactor createTransactionHash and MerkleHashBuilder

evias:
### Changed

- Refactor of Transaction.createTransactionHash for clarity and use protocol nearest transaction creation flow as seen in [TransactionExtensions](https://github.com/nemtech/catapult-server/blob/master/sdk/src/extensions/TransactionExtensions.cpp#L46)
- MerkleHashBuilder now creates hasher from instance parameters
- Fix usage of **SHA3** for transaction hashes in `Transaction.createTransactionHash` and `AggregateTransaction.calculateInnerTransactionHash


src/model/transaction/AggregateTransaction.ts
```diff
@@ -372,11 +372,23 @@ export class AggregateTransaction extends Transaction {
      * Generate inner transaction root hash (merkle tree)
      * @returns {Uint8Array}
      */
-    private calculateInnerTransactionHash(): Uint8Array {
-        const builder = new MerkleHashBuilder(SHA3Hasher.createHasher);
+    private calculateInnerTransactionHash(networkType: NetworkType): Uint8Array {
```
rg911: Seems networkType is not used at all. should it be removed?

src/model/transaction/Transaction.ts
```diff
@@ -81,29 +111,68 @@ export abstract class Transaction {
...
+     *
+     * @see https://github.com/nemtech/catapult-server/blob/master/src/catapult/model/EntityHasher.cpp#L32
+     * @see https://github.com/nemtech/catapult-server/blob/master/src/catapult/model/EntityHasher.cpp#L35
+     * @see https://github.com/nemtech/catapult-server/blob/master/sdk/src/extensions/TransactionExtensions.cpp#L46
      * @param {string} transactionPayload HexString Payload
      * @param {Array<number>} generationHashBuffer Network generation hash byte
      * @param {NetworkType} networkType Catapult network identifier
      * @returns {string} Returns Transaction Payload hash
      */
     public static createTransactionHash(transactionPayload: string, generationHashBuffer: number[], networkType: NetworkType): string {
```
rg911: again, don't think we need networkType anymore.

src/model/transaction/AggregateTransaction.ts
evias: that one can be removed

src/model/transaction/Transaction.ts
evias: this one i would rather keep as it affects CosignatureTransaction down the road ; lets wait for more clarity on keccak for removing it ?

PR #362, opened 2019-11-27
Task/g361 network in http

rg911:
- Added optional para 'NetworkType' to Http to bypass Rest calls for retrieving networkType.
- Tested on NamespaceHttp.spec.ts in e2e

### DIFF
--- a/src/infrastructure/Http.ts
+++ b/src/infrastructure/Http.ts
@@ -14,36 +14,40 @@
  * limitations under the License.
  */
 
-import {Observable, of as observableOf} from 'rxjs';
-import {map} from 'rxjs/operators';
+// tslint:disable-next-line: ordered-imports
+import {from as observableFrom, Observable, of as observableOf, throwError} from 'rxjs';
+import {catchError, map} from 'rxjs/operators';
 import {NetworkType} from '../model/blockchain/NetworkType';
-import {NetworkHttp} from './NetworkHttp';
+import { NodeRoutesApi } from './api/apis';
 import { QueryParams } from './QueryParams';
 /**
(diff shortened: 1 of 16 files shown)


src/infrastructure/Http.ts
```diff
@@ -14,36 +14,40 @@
...
-                this.networkType = networkType;
-                return networkType;
-            }));
+        if (!this.networkType) {
+            networkTypeResolve = observableFrom(new NodeRoutesApi(this.url).getNodeInfo()).pipe(
+                map(({body}) => {
+                    return body.networkIdentifier;
+                }),
+                catchError((error) =>  throwError(this.errorHandling(error))),
+            );
```
fboucquez: Every time getNetworkTypeObservable is called a new NodeRoutesApi is created and the rest endpoint is called. This is not very efficient and it may ban the server. I would move the networkTypeResolve to a read-only field and create it up in the constructor.  If cached, the user has less need for hard-coding the network type in the constructor and leaves the HTTPs to handle the network type efficiently. 

Java has an [observable.cache](url) method to cache the result in memory. 

https://github.com/nemtech/nem2-sdk-java/blob/master/sdk-okhttp-client/src/main/java/io/nem/sdk/infrastructure/okhttp/RepositoryFactoryOkHttpImpl.java#L66

Once we have the factory, the cache will work much better as the cached value will be reused between multiple Http Objects (Transaction, Account, etc...).

rg911: Can put the refactoring into a separate sdk refactor task. Also the cash would probably be better to put into the implementation e.g. client wallet etc to be able to provide more control.

Issue #357, opened 2019-11-19
Service to standardize transactions (receipts resolutions)


dgarcia360:
tldr; We could have a service to flatten transactions. E.g: Given a set of transactions announced, return them with resolved aliases.

**Why?**

Namespace alias make accounts and mosaics identifiers easily recognizable. While the feature makes announcing transactions more developer-friendly, it becomes much more complicated when having to **process past transactions**. 

For example, imagine an exchange storing deposits in a database every time they receive a transaction. They query the latest transactions every 10 minutes to process which transactions the accounts of the group have received. Some of them come with the real mosaics identifier, whereas others use the aliases. The fact is that they need to have a way to identify mosaics in a transaction since they cannot trust in the latest alias resolution since the link may vary over time.

While exchanges [can retrieve the real identifier behind the aliased assets](https://nemtech.github.io/guides/blockchain/getting-the-mosaic-identifier-behind-a-namespace-with-receipts.html#getting-into-some-code), they need first to learn about aliases and receipts, **raising the learning curve**. Note that this problem does not apply only to exchanges (just an example), but to any other use case requiring to process transactions.

**Next steps**

- Decide if we want to treat this in the SDK level or as a separate library.
- Define the service.

dgarcia360:
A possible solution. Under Transaction, Namespace, or Receipt Service:

```
  /**
    * @param transactions List of transactions.
    * @returns Observable<Transaction[]>
    */
     public resolveAliases(transations: transactions): Observable<Transaction[]>;
```

You can see [here](https://nemtech.github.io/guides/blockchain/getting-the-mosaic-identifier-behind-a-namespace-with-receipts.html#getting-into-some-code) how to get the resolution behind an aliased mosaicId.

rg911:
> A possible solution. Under Transaction, Namespace, or Receipt Service:
> 
> ```
>   /**
>     * @param transactions List of transactions.
>     * @returns Observable<Transaction[]>
>     */
>      public resolveAliases(transations: transactions): Observable<Transaction[]>;
> ```
> 
> You can see [here](https://nemtech.github.io/guides/blockchain/getting-the-mosaic-identifier-behind-a-namespace-with-receipts.html#getting-into-some-code) how to get the resolution behind an aliased mosaicId.

Not a fan of this approach (sorry mate). Would it be better to add flag in those transactions (contain alias) to be able to switch between resolved and unresolved (probably will reduce a lot duplicated codes and efforts)?

[...]

PR #363, opened 2019-11-30
Task/g357 standardize transaction service

rg911:
- Added service to resolve alias(es) for transactions from block receipts. 
- Fixed unresolvedAddress bug in `ResolutionStatement`
- Added e2e tests

**Please note that there's discussion on NEM2 slack channel regarding the `PrimaryId` value in `ResolutionStatement` as if `Harvest_Fee` receipt exists on the block, the transaction index is less `1` against the `Resolution entry primaryId`. Solution here may change upon investigation results on catapult-server**

Issue: #357


src/service/TransactionService.ts
```diff
@@ -0,0 +1,93 @@
...
+     * Constructor
+     * @param url Base catapult-rest url
+     */
+    constructor(url: string) {
+        this.transactionHttp = new TransactionHttp(url);
+        this.receiptHttp = new ReceiptHttp(url);
+    }
+
+    /**
+     * @internal
```
evias: could this method be exposed outside too?

src/service/TransactionService.ts
```diff
@@ -0,0 +1,94 @@
...
+        const resolutionStatement = (resolutionType === ResolutionType.Address ? statement.addressResolutionStatements :
+            statement.mosaicResolutionStatements).find((resolution) => resolution.height.toString() === height &&
+                (resolution.unresolved as NamespaceId).equals(unresolved));
+
+        if (!resolutionStatement) {
+            throw new Error('No resolution statement found');
+        }
+        // source (0,0) is reserved for blocks, source (n, 0) is for txes, where n is 1-based index
+        const resolutionEntry = resolutionStatement.resolutionEntries
+            .find((entry) => entry.source.primaryId ===
```
dgarcia360: ``entry.source.primaryId ===``  might not work always if there is more than one transaction in the block using the same alias (and the alias does not change within the block). I believe there is only one resolution statement if this is repeated.

**Example**
_Block:_
Height: 5
Transactions:
* 1 - Bob -> 1 random.token -> Alice
* 2- Carol -> 1 random.token > Denis

_Receipts for 5_
* mosaicResolutionStatement (random.token -> ID starting at 1)

src/service/TransactionService.ts
```diff
@@ -0,0 +1,94 @@
...
+     * Extract resolved address | mosaic from block receipt
+     * @param resolutionType Resolution type: Address / Mosaic
+     * @param unresolved Unresolved address / mosaicId
+     * @param statement Block receipt statement
+     * @param transactionIndex Transaction index
+     * @param height Transaction height
+     * @param aggregateTransactionIndex Transaction index for aggregate
+     * @returns {MosaicId | Address}
+     */
+    public static getResolvedFromReceipt(resolutionType: ResolutionType,
```
fboucquez: I'll split this method in 2, getResolvedAddressFromRecipient and getResolvedMosaicIdFromRecipient

src/model/transaction/AccountAddressRestrictionTransaction.ts
```diff
@@ -190,4 +195,59 @@ export class AccountAddressRestrictionTransaction extends Transaction {
...
         return transactionBuilder.serialize();
     }
+
+    /**
+     * @internal
+     * @param receiptHttp ReceiptHttp
+     * @param aggregateTransactionIndex Transaction index for aggregated transaction
+     * @returns {Observable<AccountAddressRestrictionTransaction>}
+     */
+    resolveAliases(receiptHttp: ReceiptHttp, aggregateTransactionIndex?: number): Observable<AccountAddressRestrictionTransaction> {
```
fboucquez: I am not fan of the double dispatch. It gets the Transaction model too complicated with internal methods the user doesn't really care. I would create a switch/handers statement in the TransactionService (especially when there is a static call  TransactionService.getResolvedFromReceipt) and remove resolveAliases from the transaction hierarchy.

I would stop using the Http implementations and start using the Repository Interface, in this case, ReceiptRepository

rg911: Do you mean to put all codes in to one class here? It's another option agree but the single class  can be complicated and main thing here is you will lose the single responsibility for Transaction to resolve alias. 

Regrading the interface, ts doesn't work like JAVA or C# which you can use unity or spring, There's no DI in TS SDK, so  resolving the interface is not easy

src/service/TransactionService.ts
rg911: Can do but will probably create blocks of duplicated codes ??

fboucquez: you can have a third method that does the filtering primary/secondary filtering. I'll push the java example

src/model/transaction/AggregateTransaction.ts
```diff
@@ -400,4 +403,32 @@ export class AggregateTransaction extends Transaction {
...
+     * @param receiptHttp ReceiptHttp
+     * @returns {Observable<AggregateTransaction>}
+     */
+    resolveAliases(receiptHttp: ReceiptHttp): Observable<AggregateTransaction> {
+        const transactionInfo = this.checkTransactionHeightAndIndex();
+        return from(this.innerTransactions).pipe(
+            mergeMap((transaction) => transaction.resolveAliases(receiptHttp, transactionInfo.index)),
+            map((transaction) => transaction as InnerTransaction),
+            toArray(),
+        ).pipe(
```
fboucquez: Should aggregateTransactionIndex be provided to resolve aliases here? aggregateTransactionIndex seems like it's never provided.

src/model/transaction/AccountAddressRestrictionTransaction.ts
fboucquez: Not about DI, it's just about using interfaces, you can do

`resolveAliases(recipientRepository: RecipientRepository, aggregateTransactionIndex?: number)`

src/model/receipt/ResolutionStatement.ts
```diff
@@ -102,38 +102,70 @@ export class ResolutionStatement {
...
-            /*
-            If no most recent secondaryId matched transaction index, find previous resolution entry (most recent).
-            */
+        /*
+        If no most recent secondaryId matched transaction index, find previous resolution entry (most recent).
+        This means the resolution entry for the specific innter transaction (inside Aggregate) /
+        was generated previously outside the aggregate. It should return the previous entry (privious primarId)
+
+        e.g. Given:
+        Entries: [{P:1, S:0}, {P:2, S:0}, {P5:, S:6}]
```
dgarcia360: P5: --> P:5

src/model/receipt/ResolutionStatement.ts
```diff
@@ -102,38 +102,70 @@ export class ResolutionStatement {
...
+        // When transaction index matches a primary id, get the most recent secondaryId (resolvedPrimaryId can only <= primaryId)
+        const resolvedSecondaryId = this.getMaxSecondaryIdByPrimaryIdAndSecondaryId(resolvedPrimaryId, secondaryId);
 
-        if (resolvedSecondaryId === 0 && secondaryId !== resolvedSecondaryId) {
-            /*
-            If no most recent secondaryId matched transaction index, find previous resolution entry (most recent).
-            */
+        /*
+        If no most recent secondaryId matched transaction index, find previous resolution entry (most recent).
+        This means the resolution entry for the specific innter transaction (inside Aggregate) /
```
dgarcia360: inter --> inner

src/model/receipt/ResolutionStatement.ts
```diff
@@ -102,38 +102,70 @@ export class ResolutionStatement {
...
+            return this.resolutionEntries.find((entry) => entry.source.primaryId === lastPrimaryId &&
+                entry.source.secondaryId === this.getMaxSecondaryIdByPrimaryId(lastPrimaryId));
         }
-        // All matched
+
+        /*
+        Found a matched resolution entry on both primaryId and secondaryId
+
+        e.g. Given:
+        Entries: [{P:1, S:0}, {P:2, S:0}, {P5:, S:6}]
```
dgarcia360: P5: --> P:5

src/model/receipt/ResolutionStatement.ts
```diff
@@ -102,38 +102,70 @@ export class ResolutionStatement {
         */
         const resolvedPrimaryId = this.getMaxAvailablePrimaryId(primaryId);
 
+        /*
+        If no primaryId found, it means there's no resolution entry available for the process. Invalid entry.
+
+        e.g. Given:
+        Entries: [{P:2, S:0}, {P5:, S:6}]
```
dgarcia360: P5: -> P:5

src/model/receipt/ResolutionStatement.ts
```diff
@@ -102,38 +102,70 @@ export class ResolutionStatement {
...
+        */
         if (resolvedPrimaryId === 0) {
             return undefined;
         } else if (primaryId > resolvedPrimaryId) {
             /*
             If the transaction index is greater than the overall most recent source primary id.
             Use the most recent resolution entry (Max.PrimaryId + Max.SecondaryId)
+
+            e.g. Given:
+            Entries: [{P:1, S:0}, {P:2, S:0}, {P:4, S:2}, {P:4, S:4} {P7:, S:6}]
```
dgarcia360: P:7

src/model/receipt/ResolutionStatement.ts
```diff
@@ -102,38 +102,70 @@ export class ResolutionStatement {
...
+            It should return Entry: {P:2, S:0}
             */
             return this.resolutionEntries
                 .find((entry) => entry.source.primaryId === resolvedPrimaryId &&
                     entry.source.secondaryId === this.getMaxSecondaryIdByPrimaryId(resolvedPrimaryId));
         }
 
-        // When transaction index matches a primary id, get the most recent secondaryId
-        const resolvedSecondaryId = this.getMaxSecondaryIdByPrimaryId(resolvedPrimaryId);
+        // When transaction index matches a primary id, get the most recent secondaryId (resolvedPrimaryId can only <= primaryId)
```
dgarcia360: primary id -> primaryId

src/model/receipt/ResolutionStatement.ts
dgarcia360: innter -> inner

src/model/receipt/ResolutionStatement.ts
```diff
@@ -102,38 +102,70 @@ export class ResolutionStatement {
...
+        const resolvedSecondaryId = this.getMaxSecondaryIdByPrimaryIdAndSecondaryId(resolvedPrimaryId, secondaryId);
 
-        if (resolvedSecondaryId === 0 && secondaryId !== resolvedSecondaryId) {
-            /*
-            If no most recent secondaryId matched transaction index, find previous resolution entry (most recent).
-            */
+        /*
+        If no most recent secondaryId matched transaction index, find previous resolution entry (most recent).
+        This means the resolution entry for the specific innter transaction (inside Aggregate) /
+        was generated previously outside the aggregate. It should return the previous entry (privious primarId)
```
dgarcia360: privious -> previous
PrimarId -> primaryId

src/model/receipt/ResolutionStatement.ts
```diff
@@ -102,38 +102,70 @@ export class ResolutionStatement {
...
             If the transaction index is greater than the overall most recent source primary id.
             Use the most recent resolution entry (Max.PrimaryId + Max.SecondaryId)
+
+            e.g. Given:
+            Entries: [{P:1, S:0}, {P:2, S:0}, {P:4, S:2}, {P:4, S:4} {P7:, S:6}]
+            Transaction: [Inx:5(0+1), AggInx:0]
+            It should return Entry: {P:3, S:4}
+
+            e.g. Given:
+            Entries: [{P:1, S:0}, {P:2, S:0}, {P:4, S:2}, {P:4, S:4}, {P7:, S:6}]
```
dgarcia360: P7: -> P:7

src/model/receipt/ResolutionStatement.ts
dgarcia360: P5->P:5

src/model/transaction/AccountAddressRestrictionTransaction.ts
rg911: How do you find the correct implementation of the injected interface?

src/model/transaction/AggregateTransaction.ts
rg911: outdated

src/service/TransactionService.ts
rg911: outdated

Issue #353, opened 2019-11-15
MaxFee improvement


dgarcia360:
1. Expected behavior and actual behavior.

By definition, the effective fee is calculated as ``effectiveFee=transaction::size∗block::feeMultiplier``. 

For that reason, we need to know the transaction's size before announcing it to the network and multiply it by an estimated maximum value to get an estimated maximum value we are willing to pay.

Although it's possible to retrieve the transaction's size, this is only possible once the fee has been defined. This leads to having to create the ``Transaction`` object twice: the first time to get the ``Transaction`` object size, and the second one to add the correspondent ``maxFee`` value.

Let's discuss in this issue how we can solve this problem. Ideally, it should be possible to define an arbitrary ``maxFee``, but just to pass a multiplier to not to have to know the transaction' size beforehand.

2. Steps to reproduce the problem.

To set a reasonable ``maxFee``, we need to define the same tx twice:

````
const size = SecretProofTransaction.create(
    Deadline.create(),
    HashType.Op_Sha3_256,
    secret,
    bobPrivateChainAccount.address,
    proof,
    NetworkType.MIJIN).size;

const maxFee = UInt64.fromUint(size * 2);

const transaction = const size = SecretProofTransaction.create(
    Deadline.create(),
    HashType.Op_Sha3_256,
    secret,
    bobPrivateChainAccount.address,
    proof,
    NetworkType.MIJIN,
    maxFee);
````

**Comment by fboucquez** [comment]
Good catch David.  In java we have the factories. We can calculate the fee
before the transaction is created and the user can update the max fee
before getting the final transaction. I'll work on examples.

**Comment by fboucquez** [comment]
Good catch David.  In java we have the factories. We can calculate the fee
before the transaction is created and the user can update the max fee
before getting the final transaction. 

Example:

```
TransferTransactionFactory factory = TransferTransactionFactory.create(
       getNetworkType(),
       recipient.getAddress(),
       Collections.singletonList(NetworkCurrencyMosaic.createAbsolute(amount)),
       new PlainMessage("Some Message")
);
factory.maxFee(BigInteger.valueOf(2 * factory.getSize()));
TransferTransaction transferTransaction = factory.build();
```

evias:
This problem is general to any fields in transactions.

I have elevated the issue with for example `addTransactions` and `addCosignatures` in AggregateTransaction.

Talking about `maxFee` is the same, making it editable outside the constructor fixes the problem.

[...]

PR #369, opened 2019-12-07
Task/g353 max fee

rg911:
Added: `setMaxFee` factory method in transaction to take `feeMultiplier`
Optimised: Transaction.resolveAlias() using object assign
Added: unit tests for Transaction.resolveAlias and Transaction.setMaxFee

Issue: #353

### DIFF
--- a/src/model/transaction/TransferTransaction.ts
+++ b/src/model/transaction/TransferTransaction.ts
@@ -295,7 +295,7 @@ export class TransferTransaction extends Transaction {
      * @param feeMultiplier The fee multiplier
      * @returns {TransferTransaction}
      */
-    public setMaxFee(multiplier: number): TransferTransaction {
-        return Object.assign({__proto__: Object.getPrototypeOf(this)}, this, {maxFee: UInt64.fromUint(this.size * multiplier)});
+    public setMaxFee(feeMultiplier: number): TransferTransaction {
+        return Object.assign({__proto__: Object.getPrototypeOf(this)}, this, {maxFee: UInt64.fromUint(this.size * feeMultiplier)});
     }
 }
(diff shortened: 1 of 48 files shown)


src/model/transaction/TransferTransaction.ts
```diff
@@ -280,22 +280,13 @@ export class TransferTransaction extends Transaction {
...
-            this.signature,
-            this.signer,
-            this.transactionInfo,
-        );
+        return Object.assign({__proto__: Object.getPrototypeOf(this)}, this,
+            {
+                recipientAddress: statement.resolveAddress(this.recipientAddress,
+                    transactionInfo.height.toString(), transactionInfo.index, aggregateTransactionIndex),
+                mosaics: this.mosaics.map((mosaic) => statement.resolveMosaic(mosaic, transactionInfo.height.toString(),
+                    transactionInfo.index, aggregateTransactionIndex))});
```
evias: Can this be changed for :

```
return {}.extend(this, {...}) as TransferTransaction
```

-- 

I believe with extend, it respect code clarity guidelines as with your `Object.assign` it adds the knowledge of _prototype_ that we may avoid.

What you are trying to do here: 

- remove the initialisation of objects by passing each fields (`return new TransferTransaction(fields here)`)
- return an **extended** object
- return an object **type cast** as TransferTransaction

I believe for this matter, the `{}.extend() as SpecializedClass` adds more clarity and does same job ; Would need small test to make sure, just wanted this to be discussed.

rg911: They both should do the merge. but Object.assign can return undefiled if the object value merged is undefined. but extend might pick the original (here in our case the unresolved) value if new value undefined. 
Don't think it will make much differences here, as the resolution process should throw error for undefined anyway, but IMO, returning undefined can be better as it avoids silent fail just in case

fboucquez: Should you be using the object spread operator?

https://basarat.gitbooks.io/typescript/docs/spread-operator.html

PR #375, opened 2019-12-09
v0.16.0 release

rg911:
## [0.16.0] - 09-Dec-2019

**Milestone**: Fushicho.3

- Added epochAdjustment introduced in catapult-server v0.9.1.1 (Fushicho.3)
- Added `setMaxFee` to `Transaction` which calculate max transaction fee using `feeMultiplier * transaction.size`.
- Optimised `resolveAlias` implementation in transaction using object spread operator.
- Fixed couple of bugs in `blockHttp` and unit tests
- Fixed travis-CI issue
- Fixed `coveralls` always return 100% issue

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,6 +12,7 @@
     "stripInternal": true,
     "outDir": "dist",
     "allowJs": false,
+    "skipLibCheck": true,
     "lib": [
       "es2017",
       "dom"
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "stripInternal": true,
     "outDir": "dist",
     "allowJs": false,
+    "skipLibCheck": true,
```
dgarcia360: What's the rationale for this change?

rg911: It looks like travis was always trying to install the latest typescript 3.7.3 where we are only on 2.7.2 for now.

PR #391, opened 2019-12-18
Fixed #390 Spread operator does not assign super class methods

### LINKED ISSUE #390: Transaction#setMaxFee() returns a broken transaction.

`Transaction#setMaxFee()` should returns callee object type and it should be signable.

ex)
TransferTransaction#setMaxFee() => TransferTransaction
AggregateTransaction#setMaxFee() => AggregateTransaction

but it returns simple Object and lost attributes, just has `maxFee`.

see this reproduce code.
https://runkit.com/nizveyl/5df9d11ce7c3d5001980fabe

3. Specifications like the version of the project, operating system, or hardware.

nem2-sdk: v0.16.0

rg911:
Fixed #390

### DIFF
--- a/test/infrastructure/RepositoryFactory.spec.ts
+++ b/test/infrastructure/RepositoryFactory.spec.ts
@@ -13,19 +13,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { expect } from "chai";
-import { RepositoryFactoryHttp } from "../../src/infrastructure/RepositoryFactoryHttp";
-import { BlockRepository } from "../../src/infrastructure/BlockRepository";
-import { instance, mock, when } from "ts-mockito";
-import { BlockInfo } from "../../src/model/blockchain/BlockInfo";
+import { expect } from 'chai';
 import { of as observableOf } from 'rxjs';
-import { map } from "rxjs/operators";
-import { NetworkRepository } from "../../src/infrastructure/NetworkRepository";
-import { NetworkType } from "../../src/model/blockchain/NetworkType";
+import { map } from 'rxjs/operators';
+import { instance, mock, when } from 'ts-mockito';
+import { BlockRepository } from '../../src/infrastructure/BlockRepository';
+import { NetworkRepository } from '../../src/infrastructure/NetworkRepository';
+import { RepositoryFactoryHttp } from '../../src/infrastructure/RepositoryFactoryHttp';
+import { BlockInfo } from '../../src/model/blockchain/BlockInfo';
+import { NetworkType } from '../../src/model/blockchain/NetworkType';
 
 describe('RepositoryFactory', () => {
     it('Should create repositories', () => {
-        const repositoryFactory = new RepositoryFactoryHttp("http://localhost:3000");
+        const repositoryFactory = new RepositoryFactoryHttp('http://localhost:3000');
 
         expect(repositoryFactory.createBlockRepository()).to.be.not.null;
         expect(repositoryFactory.createNetworkRepository()).to.be.not.null;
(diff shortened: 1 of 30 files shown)


fboucquez:
Hi @rg911 , I'm sorry for the late review. 

I think the assign operation is a bit complicated a verbose
````
Object.assign({__proto__: Object.getPrototypeOf(someObject)}, someObject, anotherObject)
````

What I would do is hide that into a static helper method (`DtoMapping.ts` maybe) and call that method where you need to do the cloning and copy ops.

For example:
````
 /**
     * This method creates a copy of the first object adding the attributes of the second object.
     * @param object the object to be cloned
     * @param attributes the extra attributes to be added to the object.
     * @returns a copy of the first object with the new attributes added.
     */
    public static assign<T>(object: T, attributes: any): T {
        return Object.assign({__proto__: Object.getPrototypeOf(object)}, object, attributes);
    }

````

and change the callers like:
````
 /**
     * Set transaction maxFee using fee multiplier
     * @param feeMultiplier The fee multiplier
     * @returns {TransferTransaction}
     */
    public setMaxFee(feeMultiplier: number): Transaction {
        return DtoMapping.assign(this, {maxFee: UInt64.fromUint(this.size * feeMultiplier)});
    }
````
Benefit: 
1) Easier to understand.
2) Quite used operation
3) If there is a better way (or not buggy, sorry for the spread operation suggestion I missed that ts secondary effect) and you need to change it, you would change it in one place only.


What do you think?

**Comment by rg911** [comment]
> Hi @rg911 , I'm sorry for the late review.
> 
> I think the assign operation is a bit complicated and verbose
> 
> ```
> Object.assign({__proto__: Object.getPrototypeOf(someObject)}, someObject, anotherObject)
> ```
> 
> What I would do is hide that into a static helper method (`DtoMapping.ts` maybe) and call that method where you need to do the cloning and copy ops.
> 
> For example:
> 
> ```
>  /**
>      * This method creates a copy of the first object adding the attributes of the second object.
>      * @param object the object to be cloned
>      * @param attributes the extra attributes to be added to the object.
>      * @returns a copy of the first object with the new attributes added.
>      */
>     public static assign<T>(object: T, attributes: any): T {
>         return Object.assign({__proto__: Object.getPrototypeOf(object)}, object, attributes);
>     }
> ```
> 
> and change the callers like:
> 
> ```
>  /**
>      * Set transaction maxFee using fee multiplier
>      * @param feeMultiplier The fee multiplier
>      * @returns {TransferTransaction}
>      */
>     public setMaxFee(feeMultiplier: number): Transaction {
>         return DtoMapping.assign(this, {maxFee: UInt64.fromUint(this.size * feeMultiplier)});
>     }
> ```
> 
> Benefits:
> 
> 1. Easier to understand.
> 2. Quite used operation
> 3. If there is a better way (or not buggy, sorry for the spread operation suggestion I missed that ts secondary effect) and you need to change it, you would change it in one place only.
> 
> What do you think?

Fixed

File: test/infrastructure/RepositoryFactory.spec.ts [hunk 1]
```diff
@@ -63,10 +63,10 @@ describe('RepositoryFactory', () => {
...
             }
-        })("http://localhost:3000");
+        })('http://localhost:3000');
 
         expect(counter).to.be.equals(0);
-        repositoryFactory.getGenerationHash().subscribe(gh => {
+        repositoryFactory.getGenerationHash().subscribe((gh) => {
             expect(counter).to.be.equals(1);
             expect(gh).to.be.equals('aaaa');
             repositoryFactory.getGenerationHash().subscribe(gh => {
```
fboucquez: Thanks, @rg911 for the cleanup. Question, why the braces in arrow function's parameter when it is not necessary?

PR #396, opened 2019-12-21
create SimpleWallet from DTO, fixes #392

### LINKED ISSUE #392: Feature request: Create a SimpleWallet from DTO
The goal of using a `SimpleWallet `is to safely store a private key.

However, when you retrieve a `SimpleWallet `from a stringified object, it's a little painful to instantiate.

I think the SDK should provide a static method to do so, or maybe modify the constructor to be able to do `new SimpleWallet(simpleWalletWithoutPrototype)`

decentraliser:
static method to instantiate a SimpleWallet that has no prototype


src/infrastructure/model/simpleWalletDTO.ts
```diff
@@ -0,0 +1,44 @@
...
+ * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
+ * https://openapi-generator.tech
+ * Do not edit the class manually.
+ */
+
+/**
+ * Used to instantiate a SimpleWallet
+ */
+export interface ISimpleWalletDTO {
+ name: string;
```
rg911: Indentation doesn't looks quite right here? Should be `tab` I think.
Also, I would create a new folder `wallet` under infrastructure root as `model` is mainly used for serving Rest API and get constantly trashed and recreated from OpenAPI generated code.

PR #404, opened 2019-12-26
export simpleWalletDTO

decentraliser:
sorry about this one... 
sorted the exports along the way


evias:
is there no other DTOs that need to be exported ?

Kind of a bad thing to do to expose *one specific* class in there, I would prefer it to expose a DTO submodule (with any required DTOs)

decentraliser:
The other DTOs are DTO that are used for typing REST API results, I separated this one from these upon demand of @rg911  for not impacting OpenAPI documentation automatic generation

fboucquez:
I agree with Steve. I think this PR belongs to the wallet project. I don't think other apps may care about it.

decentraliser:
https://github.com/nemtech/nem2-cli/blob/3eb348659ed9043e395b448cd76647a52b0a757e/src/model/profile.ts#L18-L42

This is the usecase for it

fboucquez:
Should the wallet be part of the NEM cli?

decentraliser:
In the end, it is just a small detail.

Maybe I don't know what the "wallet project" is?

In my opinion, this should be provided by the SDK, here is my reasoning:
- The SDK provides the SimpleWallet class, with handy methods to encrypt and decrypt a private key
- the only use case for a simple wallet is to be stored and retrieved
- if the SDK provides a SimpleWallet method, it should cover its full lifecycle

it's not only about the CLI, but this is also used in the desktop wallet, the browser extension wallet, and I guess in the mobile wallet as well

rg911:
> In the end, it is just a small detail.
> 
> Maybe I don't know what the "wallet project" is?
> 
> In my opinion, this should be provided by the SDK, here is my reasoning:
> 
> * The SDK provides the SimpleWallet class, with handy methods to encrypt and decrypt a private key
> * the only use case for a simple wallet is to be stored and retrieved
> * if the SDK provides a SimpleWallet method, it should cover its full lifecycle
> 
> it's not only about the CLI, but this is also used in the desktop wallet, the browser extension wallet, and I guess in the mobile wallet as well

Okay, in that case, like @evias suggested, could a submodule be created? and we already have a `wallet` folder in `model`, can it be put with other wallet classes together?

PR #406, opened 2019-12-27
TransactionHelper deserialization

fboucquez:
This PR is the SDK improvement from creating a catbuffer library via the generator. PR https://github.com/nemtech/catbuffer-generators/pull/67

If you are happy with the solution, we can release a proper catbuffer version (like 2.0.0) and update the current catbuffer lib version which is 0.0.4-SNAPSHOT

1. Removed catbuffer code
2. Added catbuffer dependency
3. Improved serialization to use TransactionHelper and builders when they are already parsed.
4. Added tslint task to npm (codestyle fixes still need fixing. Then we can enable Travis to do the checking)


src/core/utils/TransactionMapping.ts
```diff
@@ -33,11 +32,9 @@ export class TransactionMapping {
...
      * Create transaction class from payload binary.
      * @param {string} payload The transaction binary payload
-     * @param {Boolean} isEmbedded Is embedded transaction (Default: false)
-     * @returns {Transaction | InnerTransaction} The transaction class.
+     * @returns {Transaction} The transaction class.
      */
-    public static createFromPayload(payload: string,
-                                    isEmbedded = false): Transaction | InnerTransaction {
-        return CreateTransactionFromPayload(payload, isEmbedded);
+    public static createFromPayload(payload: string): Transaction {
```
fboucquez: Hi @rg911 

I believe this is the only public API method that breaks. The removal of the optional isEmbedded. Alhotugh I don't think clients would be calling this method with isEmbedded = true, I can add the isEmbedded flag back and return InnerTransaction like the old signature. What do you think? Happy to split the PR if you still find it necessary.
This PR doesn't remove the InnerTransaction.

rg911: I tend to agree but am not 100% convinced using `Transaction` as `EmbeddedTransaction`. Code will work but the concept of EmbeddedTransaction becomes a confusing.  @evias @dgarcia360 

But simply from PR point of view, it should be split

fboucquez: Fair enough, I'll create a PR with just the catbuffer library and the current master code (I'll remove catbuffer code, include the package and just update the imports)

I'll keep this PR for the TransactonHelper usage. Should we remove the InnerTransaction here?

@evias, we are noticing that InnertTransaciton type doesn't add anything special to Transaction. The only added attribute signer is already in Transaction superclass. We want to remove it simplifying the models. It's a small breaking change. You would need to change any reference from InnerTransaction to Transaction.

fboucquez:
Created https://github.com/nemtech/nem2-sdk-typescript-javascript/pull/408 that only includes the catbuffer library. 

This PR will just include the TransactionHelper improvement and probably the removal of InnerTransaction once users are happy.

src/core/utils/TransactionMapping.ts
evias: not sure that many people would be using `InnerTransaction` much vs specialized `TransferTransaction` with call to toAggregate().

fboucquez: Thanks @evias , 

1) would you confirm I can remove InnerTransaction?

2) Is the simplier TransactionMapping method ok?

`public static createFromPayload(payload: string): Transaction`

evias: 1. I couldnt confirm the removal of InnerTransaction completely because this class may also be used when trying to read things from chain inside aggregates and I cant speak for all clients that this is required/not required.

2. Take following two payloads:

2.1: Simple transfer transaction (extends `Transaction`)
 
```
CD00000000000000A775C5ACE3D7167447629EB7BB535D87E6252D46B550E4E98EBF16B3815F60BC47080010FD3C3817620C49B2E50AB3562A88968A0244BF17DDDB268A2CFF220130664BF316E6CC84B668D283F15F783724640FD4CFD1229D7D9C7604EDF46805000000000198544140420F00000000004A941B050100000098DAD0DFE1E7049C9606A7EE037B528493075B1C7F540161E3011D000000000029CF5FD941AD25D501000000000000000053696D706C65207472616E73666572206F662031206E656D2E78656D
```

2.2:  Aggregate transaction with one transfer transaction (extends `Transaction`)

```
1001000000000000D29C33F6983DA4DF650E3C5E638494C8620DF7A28C73E375917C92958B0A139FA4EE95F7175B7590811E0440CEF418D85D440C323D70571DC514DFC4A3CD260830664BF316E6CC84B668D283F15F783724640FD4CFD1229D7D9C7604EDF46805000000000198414140420F00000000005A831C05010000003E04C29404A49C65964BE1FFBB7E7EB7918D6211430EE8F8531BF5995175682F6800000000000000610000000000000030664BF316E6CC84B668D283F15F783724640FD4CFD1229D7D9C7604EDF46805000000000198544198DAD0DFE1E7049C9606A7EE037B528493075B1C7F540161E30101000000000029CF5FD941AD25D501000000000000000000000000000000
```

2.3: Embedded transaction content _of one transfer transaction_ (extends `InnerTransaction`)

```
98DAD0DFE1E7049C9606A7EE037B528493075B1C7F540161E30101000000000029CF5FD941AD25D501000000000000000000000000000000
```

One thing I can think about is that _removing the isEmbedded flag_ in the case of aggregate transactions, a client could share only the embedded payload (like in 2.3) so that other clients have correct content (but its not defined which key should sign). As a compromise, you could also provide with a `createFromInnerPayload` method.

fboucquez: 1. I would remove it, it won't be a big change for the customers as it just changing the class name. Currently, the downcasting is pretty odd at least
2. I'll add the createFromInnerPayload in TransferTransaction

rg911: regarding all these changes mentioned, why can't it just keep as it is now?

fboucquez: I'm fine either way. 2 methods or 1 method with an if. I guess keeping the same signature won't break existing customers

fboucquez: Removed InnerTransaction. Code is simpler.
Added isEmbedded back to the createFromPayload
Added unit test

PR #445, opened 2020-02-12
Key derivation using sha512

rg911:
This PR introduced quite a few `Breaking changes`:
- All Key derivation and signing hash are now using `SHA512`. Means all KeyPairs generated and Addresses derived from publicKey are now having different hex values.
- Cleared the bond between `NetworkType` and `SigningSchema`. `SigningSchema` has now been removed from the SDK. There's no `Keccak / SHA3` switch anymore. 
- Key pair generation is now using `nacl.sign.keypair` from the standard nacl js script. 
- Transaction signing is now using `nacl.sign.detached` from the standard nacl js script. 
- Legacy catapult-nacl is still used for SharedKey derivation due to the multiplier methods is internal and not exported in the nacl js. 

This PR also:
- Removed unused legacy catapult-nacl code in `Utilities` class
- Fixed the tests
- Fixed a few lint issues


fboucquez:
Should we be case insensitive when comparing strings in places like:

- https://github.com/NEMStudios/nem2-sdk-typescript-javascript/blob/task/g441_key_derivation_sha512/src/infrastructure/Listener.ts#L235
- https://github.com/NEMStudios/nem2-sdk-typescript-javascript/blob/task/g441_key_derivation_sha512/src/infrastructure/Listener.ts#L336
- https://github.com/NEMStudios/nem2-sdk-typescript-javascript/blob/task/g441_key_derivation_sha512/src/model/transaction/Transaction.ts#L301
- https://github.com/NEMStudios/nem2-sdk-typescript-javascript/blob/task/g441_key_derivation_sha512/src/model/transaction/Transaction.ts#L318

Just wondering if some other clients create the transaction using lower case hex that then TS SDK will false-negative think it's not the  same

rg911:
> Should we be case insensitive when comparing strings in places like:
> 
> * https://github.com/NEMStudios/nem2-sdk-typescript-javascript/blob/task/g441_key_derivation_sha512/src/infrastructure/Listener.ts#L235
> * https://github.com/NEMStudios/nem2-sdk-typescript-javascript/blob/task/g441_key_derivation_sha512/src/infrastructure/Listener.ts#L336
> * https://github.com/NEMStudios/nem2-sdk-typescript-javascript/blob/task/g441_key_derivation_sha512/src/model/transaction/Transaction.ts#L301
> * https://github.com/NEMStudios/nem2-sdk-typescript-javascript/blob/task/g441_key_derivation_sha512/src/model/transaction/Transaction.ts#L318
> 
> Just wondering if some other clients create the transaction using lower case hex that then TS SDK will false-negative think it's not the same

I think it should be okay as REST will always return upper case. but would add extra security

PR #450, opened 2020-02-18
Block service

### LINKED ISSUE #449: BlockService for merkle auditing
Adding a new `BlockService` to validate a given merkle path item against the root hash at a specific block height. 

e.g. `validateInBlock(...)`

rg911:
Added `BlockService` to validate transaction and receipt hashes against in specific block

Fixed #449


src/service/BlockService.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+        const rootToCompare = merklePathItem.reduce((proofHash, pathItem) => {
+                const hasher = sha3_256.create();
+                // Left
+                if (pathItem.position === 1) {
+                    return hasher.update(Buffer.from(pathItem.hash + proofHash, 'hex')).hex();
+                } else {
+                    // Right
+                    return hasher.update(Buffer.from(proofHash + pathItem.hash, 'hex')).hex();
+                }
```
fboucquez: The position is being changed from a number to a (left/right) string. Rest has done the change in master, we are still waiting for the open api PR https://github.com/nemtech/nem2-openapi/pull/77

@dgarcia360 can we move 77 PR forward? once released we can then generate the clients and update the ts and java PR related to merkle tests

src/service/BlockService.ts
```diff
@@ -0,0 +1,87 @@
...
+     * @param leaf transaction hash
+     * @param height block height
+     */
+    public validateTransactionInBlock(leaf: string, height: UInt64): Observable<boolean> {
+        const rootHashObservable = this.blockRepository.getBlockByHeight(height);
+        const merklePathItemObservable = this.blockRepository.getMerkleTransaction(height, leaf);
+        return combineLatest(rootHashObservable, merklePathItemObservable).pipe(
+            map((combined) => this.validateInBlock(leaf, combined[1].merklePath, combined[0].blockTransactionsHash)),
+        );
+    }
```
fboucquez: When rest raises an error, like block, transaction or recipients not found, should this method catch the exception and return false or just raise the exception?

I think it should raise the error but no 100% sure. @rg911 @dgarcia360

fboucquez:
@rg911 can we add unit testing to the block service. E2E tests are great but they don't run every build and they don't increase coverage.

src/service/BlockService.ts
rg911: In fact if the transaction / receipt doesn't exist. The exception should be captured before the validation within other http repos

fboucquez: I think adding` .pipe(catchError(error => of(false)))` would do the trick after combineLatest

src/service/BlockService.ts
dgarcia360: https://github.com/nemtech/nem2-openapi/releases/tag/v0.8.4

e2e/service/BlockService.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+                expect(validationResult).to.be.true;
+            } else {
+                assert(false, `Transaction (hash: ${transactionHash}) not found`);
+            }
+        });
+    });
+
+    describe('Validate receipt', () => {
+        it('call block service', async () => {
+            const statement = await receiptRepository.getBlockReceipts(UInt64.fromUint(1)).toPromise();
```
dgarcia360: const statement**s**

e2e/service/BlockService.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+            const receipt = statement.transactionStatements[0];
```
dgarcia360: const ~receipt~ statement

src/service/BlockService.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Validate receipt hash in block
+     * @param leaf receipt hash
+     * @param height block height
+     */
+    public validateReceiptInBlock(leaf: string, height: UInt64): Observable<boolean> {
```
dgarcia360: Should the function be named ``validateStatementInBlock``? The hash passed as leaf belongs to a transaction/resolution statement, not to an individual receipt.

fboucquez:
@rg911 0.8.4-SNAPSHOT.202002181344 client has been pushed

rg911:
Rebased to #453

PR #453, opened 2020-02-18
Apply open api 0.8.4

### LINKED ISSUE #451: Adding new endpoint support based on OpenAPI definition
https://github.com/nemtech/nem2-openapi/pulls

rg911:
- Added positionEnum for merklePathItem
- Changed to use comma-separated string for transaction type filtering
- Added generationHash in NodeInfo
- Changed NetworkCurrencyMosaic -> NetworkCurrencyLocal (cat.currency)
- Added NetworkCurrencyPublic (symbol.xym)

- Rebased #450 

Fixed #451 #449


src/infrastructure/TransactionSearchCriteria.ts
```diff
@@ -0,0 +1,45 @@
...
+import { QueryParams } from './QueryParams';
+
+ /**
+  * Transaction search criteria for filtering transaction list
+  */
+export class TransactionSearchCriteria extends QueryParams {
+    /**
+     * Transaction type filter
+     */
+    public type?: string;
```
fboucquez: This should be a list of enums so the model is cleaner. Then the repository should do the csv to the param which is an implementation detail (I think there should be a way in open api to not use csv params)

rg911: I decided to remove this class instead using the existing queryParam class as implemented the setter

fboucquez: The problem with QuerayParams is that it's used for different queries that may not have a transaction type filtering.

For example:

https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/master/src/infrastructure/NamespaceRepository.ts#L63

Now they accept transaction types that aren't used

The plan is you have common pagination/sorting param in a top-level class (QueryParams although I would call it SearchCriteria) and then you have a subclass adding the custom filters (like TransactionSearchCriteria). Hopely one subclass per entity type (Account, Transaction, Namespace)

I don't like the name QueryParams. It is implementation-specific. What happens if tomorrow we send the params as a post body or we call the database directly like the e2e repositories? 

The search and pagination will have a big refactoring soon, this can be revisited then.

@dgarcia360 @evias ?

rg911: Replaced by TransactionFilter class which acts now as a separate parameter on account http

PR #457, opened 2020-02-20
Query params arguments in constructor, fixes #456

### LINKED ISSUE #456: QueryParams - Allow passing arguments to the constructor
https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/42fb1dd38bcfe7978b10059056f263753b836b93/src/infrastructure/QueryParams.ts#L51

Could we allow passing the pageSize and order as constructor arguments?

decentraliser:
fixes #456

### DIFF
--- a/src/infrastructure/QueryParams.ts
+++ b/src/infrastructure/QueryParams.ts
@@ -35,20 +35,34 @@ export class QueryParams {
      */
     public pageSize = 10;
     /**
-     * Id after which we want objects to be returned
-     */
-    public id?: string;
-    /**
      * Order of transactions.
      * DESC. Newer to older.
      * ASC. Older to newer.
      */
+
     public order: Order = Order.DESC;
 
     /**
-     * Constructor
+     * Id after which we want objects to be returned
      */
-    constructor() {
+    public id?: string;
+
+    /**
+     * Creates an instance of QueryParams.
+     * @param {{
+     *         pageSize?: number,
+     *         order?: Order,
+     *         id?: string;
+     *     }} configuration arguments
+     */
+    constructor(args: {
+        pageSize?: number,
+        order?: Order,
+        id?: string;
+    }) {
+        if (args.pageSize) this.setPageSize(args.pageSize)
+        if (args.order) this.setOrder(Order[args.order])
+        if (args.id) this.setId(args.id)
     }
 
     public setPageSize(pageSize: number): QueryParams {
--- a/src/infrastructure/QueryParams.ts
+++ b/src/infrastructure/QueryParams.ts
@@ -55,14 +55,16 @@ export class QueryParams {
      *         id?: string;
      *     }} configuration arguments
      */
-    constructor(args: {
+    constructor(args?: {
         pageSize?: number,
         order?: Order,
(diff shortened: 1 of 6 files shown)


src/infrastructure/QueryParams.ts
```diff
@@ -34,21 +34,35 @@ export class QueryParams {
...
+    /**
+     * Creates an instance of QueryParams.
+     * @param {{
+     *         pageSize?: number,
+     *         order?: Order,
+     *         id?: string;
+     *     }} configuration arguments
      */
-    constructor() {
+    constructor(args: {
```
fboucquez: What about 
````
 constructor(args?: {
        pageSize?: number,
        order?: Order,
        id?: string;
    }) {
        if (args && args.pageSize) {
            this.setPageSize(args.pageSize);
        }
        if (args && args.order) {
            this.setOrder(Order[args.order]);
        }
        if (args && args.id) {
            this.setId(args.id);
        }
    }
````

So you can have a call like

`new QueryParams() `without the emtpy {}

fboucquez:
Double-check unit tests and tslint. Travis is complaining about the missing {} as args is mandatory.

src/infrastructure/QueryParams.ts
decentraliser: oops sorry, just updated

decentraliser:
> I would probably remove the setter methods and make the fields readonly
they could be private as well
I think the setters are fine as is

fboucquez:
I would take it further and change the objects with something like
````
/**
 * The query params structure describes pagination params for requests.
 *
 * @since 1.0
 */
export class QueryParams {
    /**
     * Page size between 10 and 100, otherwise 10
     */
    public readonly pageSize: number;
    /**
     * Order of transactions.
     * DESC. Newer to older.
     * ASC. Older to newer.
     */
    public readonly order: Order;
    /**
     * Id after which we want objects to be returned
     */
    public readonly id?: string;
    /**
     * Creates an instance of QueryParams.
     * @param {{
     *         pageSize?: number,
     *         order?: Order,
     *         id?: string
     *     }} args arguments
     */
    constructor(args?: {
        pageSize?: number,
        order?: Order,
        id?: string;
    }) {
        this.pageSize = args && args.pageSize || 10;
        this.order = args && args.order || Order.DESC;
        this.id = args && args.id || undefined;
    }
}
````

Examples:

````
const params1 = new QueryParams({pageSize: 10});
const params2 = new QueryParams({id: 'someId'});
const params3 = new QueryParams({pageSize: 20, order: Order.ASC});
````

About `args.pageSize`, I would probably raise an exception if it's not in the valid 10...100 range. If not the user would expect a page size but the SDK is really asking for another size.

decentraliser:
>  I would probably raise an exception if it's not in the valid 10...100 range.
I let a setter for pageSize and made it private in order to handle this exception out of the constructor

> Would need to do the same on the TransactionFilter
> Also the PR description links to the wrong issue
Fixed both

fboucquez:
I will make the QueryParams fields readonly. This would be a future improvement as we are in the middle of the rebranding

PR #463, opened 2020-02-25
NPM Alpha deployments from master

fboucquez:
Addresses issue https://github.com/nemtech/symbol-sdk-typescript-javascript/issues/458. Please read the reason for this PR there.

Once this PR is merged, Travis will start deploying an alpha version to npm every master build (nodejs version 8 builder).

The notation is

`symbol-sdk:$version-alpha-$datetime`

For example

`symbol-sdk:0.17.2-alpha-202002251427` (tag alpha) 

Note that the alpha version uses the next release version, right now it would be `0.17.2-alpha-xxxx`. Once `0.17.2` is released, the next alpha version would be `0.17.3-alpha-xxxx`

I couldn't find any --pre option in 
https://docs.npmjs.com/cli/publish

Also, the --tag doesn't allow redeploy if the version is the same. The travis script generates a new alpha version when executed

Note: `NPM_TOKEN` needs to be provided in https://travis-ci.org/nemtech/symbol-sdk-typescript-javascript/settings.  I could use my token but my account would be the one publishing. Ideally, we should have a bot npm user.

The `alpha` version and tag name can be changed for anything. Maybe `prerelease`?

### DIFF
--- a/src/infrastructure/Http.ts
+++ b/src/infrastructure/Http.ts
@@ -79,7 +79,7 @@ export abstract class Http {
         if (error.code && error.address && error.code === 'ECONNREFUSED') {
             return new Error(`Cannot reach node: ${error.address}:${error.port}`);
         }
-        return new Error(error);
+        return error;
     }
 
     /**
--- a/src/infrastructure/Http.ts
+++ b/src/infrastructure/Http.ts
@@ -79,7 +79,10 @@ export abstract class Http {
         if (error.code && error.address && error.code === 'ECONNREFUSED') {
             return new Error(`Cannot reach node: ${error.address}:${error.port}`);
         }
-        return error;
+        if (error instanceof Error) {
+            return error;
+        }
+        return new Error(error);
     }
 
     /**
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,17 +8,22 @@ cache:
     - "node_modules"
 before_script:
   - npm run build
-script: 
+script:
   - npm run test:cov
   - npm run coveralls-report
   - npm install --global typedoc
   - typedoc --out ts-docs src
   - touch ./ts-docs/.nojekyll
 deploy:
-  provider: pages
-  skip_cleanup: true
-  local_dir: ts-docs
-  github_token: $GITHUB_TOKEN
-  on:
-    branch: master
-
+  - provider: pages
+    skip_cleanup: true
+    local_dir: ts-docs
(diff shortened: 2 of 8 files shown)


fboucquez:
In a perfect world, I would go with patch releases every master merge. I see some problems now:

1. SDK release requires some manual steps like writing down the Change Log, create GitHub releases from tags with the changelog, notify users via slack. Git tagging and npm version increase.  We need to automatize everything (which it's possible but not that simple) before allowing master releases. If not you would need to manually add a new entry into the changelog every new PR.
2. We may not want to publically release a new version if a dependency hasn't been publically released yet (like the current 0.9.3.1 compatible private releases) 
3. We may not want to increase a patch version every small PR merge like a doc fix, unit test added or some small code clean up.

In Java, we are using the special `release` branch to perform the patch releases where git gets tagged, the version gets upgraded,  release artifacts are uploaded, etc. We are still doing the manual step of writing the changelog, creating the release form a tag and slack notifications.

In java, a master merge means `snapshot` deployment, a  release merge means `release` deployment

fboucquez:
@rg911 I've tested the dependency resolution in npm

The user needs to explicitly tell npm when installing that you want a particular tag or alpha version

https://medium.com/@kevinkreuzer/publishing-a-beta-or-alpha-version-to-npm-46035b630dd7

Auto-upgrade only works on the latest released version (x.y.z), without any suffix.

So:

If you have something like in your package.json

`"symbol-openapi-typescript-node-client": "^0.8.2",`

It will bring the artifact number `0.8.5` although there are newer alpha/unreleased ones

https://www.npmjs.com/package/symbol-openapi-typescript-node-client

When we merge this, the user would need to explicitly tell npm that they one a particular alpha version. I don't think an unreleased artifact will seek into anybody that wants a release version

src/infrastructure/Http.ts
```diff
@@ -79,6 +79,9 @@ export abstract class Http {
         if (error.code && error.address && error.code === 'ECONNREFUSED') {
             return new Error(`Cannot reach node: ${error.address}:${error.port}`);
         }
+        if (error instanceof Error) {
+            return error;
+        }
         return new Error(error);
     }
```
fboucquez: If we just wrap the exception into a new one Error, the original stack trace is lost. We want the stack trace when something in our code fails, like an NPE when mapping the DTO.

PR #469, opened 2020-02-29
Added WS status error handling to TransactionService

fboucquez:
Added TransactionService Unit Tests
Improved e2e tests to use TransactionService announce
Fixed BlockService class not implementing the interface and deprecated calls

Fixes issue https://github.com/nemtech/symbol-sdk-typescript-javascript/issues/468

### DIFF
--- a/e2e/infrastructure/IntegrationTestHelper.ts
+++ b/e2e/infrastructure/IntegrationTestHelper.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { combineLatest } from 'rxjs';
-import { filter } from 'rxjs/operators';
+import { map } from 'rxjs/operators';
 import { IListener } from '../../src/infrastructure/IListener';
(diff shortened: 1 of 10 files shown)


fboucquez:
@rg911 

Due to coverage reporting a little bit lower than it should (I added unit tests around the TransactionService), I've added unit tests around NetworkHttp and NodeHttp to increase coverage. I had to change the constructor of both objects a little bit so the low level generated clients can be mocked and provided in the unit tests. If not, we won't be able to mock them. The added parameters are optional, existing clients won't be affected at all.

If you like this idea, we should start creating unit tests of the HTTP objects by mocking the low-level generated clients. This will increase our coverage, for example on these objects:

https://coveralls.io/builds/29046184/source?filename=src%2Finfrastructure%2FAccountHttp.ts
https://coveralls.io/builds/29046184/source?filename=src%2Finfrastructure%2FRestrictionAccountHttp.ts
https://coveralls.io/builds/29046184/source?filename=src%2Finfrastructure%2FMetadataHttp.ts

PR #484, opened 2020-03-06
Travis alpha releases from master

fboucquez:
Added missing .npmrc file


.npmrc.template
```diff
@@ -0,0 +1 @@
+//registry.npmjs.org/:_authToken=${NPM_TOKEN}
```
rg911: no new line end of the file

PR #489, opened 2020-03-10
Make namespace max depth configurable in idGenerator

### LINKED ISSUE #455: namespace_max_depth should not be hardcoded
Just came across the declaration of `namespace_max_depth` that blocks the creation of a `namespaceId` when it has too many parts.

the `maxNamespaceDepth `is configurable in network settings, it can be more than 3

https://github.com/nemtech/nem2-sdk-typescript-javascript/blob/40fa287fe079560a7b97ba3bbccecbe8f08e93ab/src/core/format/Utilities.ts#L97

rg911:
- Removed default namespace maxDepth.
- Removed namespace maxDepth validator from the sdk as it can be validated by the sever.
Fixed #455


rg911:
> Users and model objects don't use IdGenerator directly. They go through NamespaceMosaicIdGenerator (for example NamespaceId).
> 
> The max depth default param value hides it from the main model objects so they ended up using the default value.
> 
> We can either:
> 
> 1. Remove the default value and propagate maxDepth to the client objects.
> 2. Remove that validation to keep the API clean. Maybe provide a validate method after the namespace is created.
> 
> I don't particularly like either, do you have another option?

I would probably remove the default depth

PR #511, opened 2020-03-31
Native BigInt implementation POC

rg911:
This is a POC pull request for early review and feedback only.


fboucquez:
Hi Steve, I've checked out your PR. It's a good start. My points:

1) I can see catbuffer generated code uses number[]. Before trying to map those number[] to/from BigInt in the TS, I'll migrate the TS catbuffer-generator to use BigInt as too.  This will save you a lot of mappings operations, the catbuffer new BigInt numbers can be used as-is when creating the TS model objects. You can publish a BigInt new version/ new name of catbuffer (catbuffer-bigint?) into NPM and use that one for this task.
2) I would also try to tune the open API generator to use BigInt when possible. We are doing that in Java. This will save you a lot of mappings.
3) I wouldn't migrate a class one by one creating a BigInt version (like `TransferTransactionBigInt`). This will be time-consuming and then you would need to rename those classes back to the original name (`TransferTransaction`). I would just do a kind of "find/replace" UInt64 -> BigInt in the IDE and then fix the compilation problems. You can use some smarter regexes to for example to remove all the imports `\nimport \{ UInt64 \} .*;` -> `empty` (BigInt is native doesn't need to be imported).  You can also use regex capturing, some examples in the [Intellij doc](https://www.jetbrains.com/help/idea/tutorial-finding-and-replacing-text-using-regular-expressions.html). Intellij has also a way of converting a type of a field from one type to another. I guess you could do some of these regexes find/replace and refactorings in VStudio. Of course, find a replace it's not perfect, but it's a quick and dirty way of doing most of the refactoring.
4) You would need to call BigIntConvertUtil conversion pretty often (not that often if catbufer and clients can use BigInt).
5) I would reconsider methods that return a number[]  (for example toDto()). I would use the BigInt as-is or a String number to be compatible with the rest data. Look at this post: https://github.com/GoogleChromeLabs/jsbi/issues/30, Now I get what's the second param of JSON.stringify :)
6) We will be able to use native number operations like === or >. We wouldn't need to have methods like equals()
7) We have a pretty comprehensive set of Unit Tests to be sure the refactoring will be fine. We will notice when objects are not working as they should.
8) For some reason, Travis passes when using [Node 10](https://travis-ci.org/github/nemtech/symbol-sdk-typescript-javascript/builds/669179221?utm_source=github_status&utm_medium=notification). This is pretty good, can the TS be compatible with Node 10 too?
9) We need to remove node 8 and node 9 from Travis adding 12.

rg911:
@fboucquez good comment mate! Exactly what I am doing right now. This is sort of a code snippet for a quick preview. The next commit will be huge.

PR #522, opened 2020-04-09
LockHashAlgorithm

### LINKED ISSUE #518: The Op_Keccak_256 is removed from the lock hash algorithm type on the server side.
The SDK and server enum for lock hash does not match any more since Op_Keccak_256 was removed from the server side.

https://github.com/nemtech/symbol-sdk-typescript-javascript/blob/f6f62c0af39b058d54213ac7a797d4218a159e51/src/model/transaction/HashType.ts#L28

### LINKED ISSUE #521: Fix `kind-of` security alert

rg911:
- Added LockHashUtils
- Removed Op_Keccak_256 algorithm
- Renamed HashType to LockHashAlgorithm
- Renamed model field name to be aligned with Catbuffer
- Fixed security alert on `kind-of`

Fixed #518
Fixed #521

### DIFF
--- a/e2e/infrastructure/TransactionHttp.spec.ts
+++ b/e2e/infrastructure/TransactionHttp.spec.ts
@@ -54,7 +54,7 @@ import { CosignatureSignedTransaction } from '../../src/model/transaction/Cosign
 import { CosignatureTransaction } from '../../src/model/transaction/CosignatureTransaction';
 import { Deadline } from '../../src/model/transaction/Deadline';
 import { HashLockTransaction } from '../../src/model/transaction/HashLockTransaction';
-import { HashType } from '../../src/model/transaction/HashType';
+import { LockHashAlgorithm } from '../../src/model/transaction/LockHashAlgorithm';
 import { LinkAction } from '../../src/model/transaction/LinkAction';
 import { LockFundsTransaction } from '../../src/model/transaction/LockFundsTransaction';
 import { MosaicAddressRestrictionTransaction } from '../../src/model/transaction/MosaicAddressRestrictionTransaction';
@@ -71,6 +71,8 @@ import { TransactionType } from '../../src/model/transaction/TransactionType';
 import { TransferTransaction } from '../../src/model/transaction/TransferTransaction';
 import { UInt64 } from '../../src/model/UInt64';
 import { IntegrationTestHelper } from './IntegrationTestHelper';
+import { NetworkCurrencyPublic } from '../../src/model/mosaic/NetworkCurrencyPublic';
+import { LockHashUtils } from '../../src/core/utils/LockHashUtils';
 
 describe('TransactionHttp', () => {
     let transactionHash;
@@ -873,7 +875,7 @@ describe('TransactionHttp', () => {
                 Deadline.create(),
                 NetworkCurrencyLocal.createAbsolute(10),
                 UInt64.fromUint(100),
-                HashType.Op_Sha3_256,
+                LockHashAlgorithm.Op_Sha3_256,
                 sha3_256.create().update(Crypto.randomBytes(20)).hex(),
                 account2.address,
                 networkType, helper.maxFee,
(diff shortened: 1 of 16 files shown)


e2e/infrastructure/TransactionHttp.spec.ts
```diff
@@ -895,7 +897,7 @@ describe('TransactionHttp', () => {
                 Deadline.create(),
                 NetworkCurrencyLocal.createAbsolute(10),
```
fboucquez, line 898: We should delegate the creation of a currency mosaic to the helper so it can resolve if it's running locally or tesnet (different currency name)

Upgrade all secret lock transactions to use the new helper.

src/core/utils/LockHashUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+import { sha3_256 } from 'js-sha3';
+import { Convert } from '../format/Convert';
+
+/**
+ * Hash utilities for SecretLock hashing
+ */
+export class LockHashUtils {
```
fboucquez, line 23: We should add a few static test cases like

https://github.com/nemtech/symbol-sdk-java/blob/33e17792ce2926f601ba2b50f1982d7e5d8be030/sdk-core/src/test/java/io/nem/symbol/core/crypto/HashesTypescriptValidationTest.java#L40

Se we are sure we are hashing the same way

src/model/transaction/LockHashAlgorithm.ts
```diff
@@ -23,21 +23,25 @@
...
 }
 
-export function HashTypeLengthValidator(hashType: HashType, input: string): boolean {
+/**
+ * Validate the hash length
+ * @param hashAlgorithm Secret lock hash algorithm
+ * @param input Hashed value
+ * @returns {boolean}
+ */
+export function LockHashAlgorithmLengthValidator(hashAlgorithm: LockHashAlgorithm, input: string): boolean {
```
fboucquez, line 38: should we have a hash(hashAlgorithm: LockHashAlgorithm, input: string | Uint8Array):string that hashes the input using the right LockHashUtils (Hashers?). it's a similar switch statement

src/core/utils/LockHashUtils.ts
fboucquez, line 23: Should it be called `Hashes`?

rg911: these methods are only for the SecretLock hashes. Don't want to confuse the user with general hashes staffs

e2e/infrastructure/TransactionHttp.spec.ts
rg911: yep. thought about it already. will do it in other PR

src/model/transaction/LockHashAlgorithm.ts
rg911: It works differently from the Java. TS can do very limited in the enum class. Otherwise, have to export method like the validator. I think the static util class should be fine?

src/core/utils/LockHashUtils.ts
rg911: Not sure we need to do it here as the OP_ hashes are specific for SecretLock. Also we don't have a test vector for these. So I think minimal unit test + integration test should be enough?

src/model/transaction/LockHashAlgorithm.ts
fboucquez: maybe add the static function to LockHashUtils that receives a type and a value and does the proper switch statement

src/core/utils/LockHashUtils.ts
fboucquez: fair enough

fboucquez: We can create our own test vectors in the SDKs, if there are too many hash types, just focus on the ones used.

PR #530, opened 2020-04-13
upgrade crypto-js to v4.0.0, add AESEncryptionService and apply it to SimpleWallet, fixes  #524

### LINKED ISSUE #524: Refactor simple wallet key encryption to use standard AES
To eliminate the security vulnerability (rainbow tables for sha3 ), it is proposed to:

- `toMobileKey` to be renamed to `encrypt_pbkdf2`
- `derivePassSha`, `passwordToPrivateKey` and `encodePrivateKey` to be removed
- instead use standard AES encryption / decryption with seed

decentraliser:


src/model/wallet/SimpleWallet.ts
```diff
@@ -91,17 +87,15 @@ export class SimpleWallet extends Wallet {
                                 privateKey: string,
                                 network: NetworkType): SimpleWallet {
         // Create KeyPair from hash key
-        const keyPair = KeyPair.createKeyPairFromPrivateKeyString(privateKey);
+        const {publicKey} = KeyPair.createKeyPairFromPrivateKeyString(privateKey);
```
rg911, line 90: spaces missing over the destructuring

test/model/account/AccountNames.spec.ts
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2018 NEM
```
rg911, line 2: date

src/model/wallet/SimpleWallet.ts
```diff
@@ -38,44 +37,41 @@ export class SimpleWallet extends Wallet {
...
                 address: Address,
-                creationDate: LocalDateTime,
                 /**
-                 * The encrypted private key and information to decrypt it
+                 * The encrypted private key as Sjcl Cipher
                  */
-                public readonly encryptedPrivateKey: EncryptedPrivateKey) {
-        super(name, network, address, creationDate, 'simple_v1');
+                public readonly encryptedPrivateKey: SjclCipherEncrypted) {
+        super(name, address, 'simple_v2');
```
rg911, line 45: Should the hardcoded version still be used here? this will make the sdk depends on the simple wallet.

src/model/wallet/SimpleWallet.ts
decentraliser: This could be added as an eslint rule

fboucquez:
We have merged prettier with the new formatted code. For this merge now, you should accept your code when merging and the run `npm run style:fix` to fix the formatting.

There are some fixes in Crypto you want to keep from master. Double-check that conflict fix

src/core/crypto/AESEncryptionService.ts
```diff
@@ -0,0 +1,77 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Password } from '../../model/wallet/Password';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const CryptoJS = require('crypto-js');
+
+export class AESEncryptionService {
```
rg911, line 22: - If this service is only used by wallet I would call it something else e.g `SimpleWalletService`
- Otherwise, to make it generic, remove wallet reference (import), also considering to merge it into the Crypto.ts class to replace the existing `encrypt / decrypt` functions.

I would prefer the latter option.

fboucquez:
@decentraliser, is it possible to not use the require('crypto-js'); style?

decentraliser:
@fboucquez please refer to my comment here https://github.com/nemtech/symbol-sdk-typescript-javascript/issues/537

rg911:
> @decentraliser, is it possible to not use the require('crypto-js'); style?

Yeah, we decided to use require for now as the crypto-js typing lib was not in good shape

PR #532, opened 2020-04-13
Prettier airbnb

fboucquez:
Extension of https://github.com/nemtech/symbol-sdk-typescript-javascript/pull/531 with aribnb
I prefer 531, aribnb is too strict and it requires many changes


fboucquez:
I'm happy with current prettier and eslint

PR #539, opened 2020-04-15
maxFee in Aggregate transactions

### LINKED ISSUE #533: Aggregate bonded gets dropped due to incorrect max fee
With Aggregate bonded transaction its harder to calculate the min fee required for the transaction to get process.  The problem is due to the cosigner signatures is not included in the submission of the tx to the server.  The SDK has two proposals to fix this issue.

1) Use the multisig account graph as state by @evias.

What about using the multisig account graph to determine the _number of required_ signers ?

@Wayonb @fboucquez @rg911 http://api-01.us-west-1.symboldev.network:3000/account/TAWJ2M7BGKWGBPOUGD5NDKHYDDQ7OQD26HJMMQS2/multisig/graph

Knowing that aggregate bonded also have there own "announcement method". I can see this help because the account multisig *graph* can be iterated over to determine the needed *total* number of minApproval

_Originally posted by @evias in https://github.com/nemtech/symbol-sdk-java/issues/251#issuecomment-601109696_

This is gives an accurate min fee but has a lot of cost to figure out all the required cosigners.

2) Always calculate the max fee using the max cosigners as stated by @dgarcia360.

	One costly option for the users would be adding maxCosignaturesPerAggregate * cosignatureSize to the aggregate transaction size. Users can still pay fewer fees if they want to adjust better the cost by reducing the transaction's maxFee multiplier.

		MaxCosignaturesPerAggregate can be retrieved from https://github.com/nemtech/catapult-server/blob/master/resources/config-network.properties#L47

		This change requires to have a configuration parameter on the SDK to define which is the default maxCosingaturesPerAggregate since it might vary per network. The value could also be retrieved from REST if not specified.

rg911:
- Added setMaxFeeForAggregate
- Changed setMaxFee for stand alone tx only
- Added getMaxCosignatures
- Added getNetworkMaxCosignaturesPerAggregate

Fixed #533

### DIFF
--- a/src/service/AggregateTransactionService.ts
+++ b/src/service/AggregateTransactionService.ts
@@ -24,16 +24,24 @@ import { InnerTransaction } from '../model/transaction/InnerTransaction';
 import { MultisigAccountModificationTransaction } from '../model/transaction/MultisigAccountModificationTransaction';
 import { SignedTransaction } from '../model/transaction/SignedTransaction';
 import { TransactionType } from '../model/transaction/TransactionType';
+import { Address } from '../model/account/Address';
+import { RepositoryFactory } from '../infrastructure/RepositoryFactory';
+import { NetworkRepository } from '../infrastructure/NetworkRepository';
 
 /**
  * Aggregated Transaction service
  */
 export class AggregateTransactionService {
+    private readonly multisigRepository: MultisigRepository;
+    private readonly networkRepository: NetworkRepository;
     /**
      * Constructor
      * @param multisigRepository
      */
-    constructor(private readonly multisigRepository: MultisigRepository) {}
+    constructor(public readonly repositoryFactory: RepositoryFactory) {
+        this.multisigRepository = repositoryFactory.createMultisigRepository();
+        this.networkRepository = repositoryFactory.createNetworkRepository();
+    }
 
     /**
      * Check if an aggregate complete transaction has all cosignatories attached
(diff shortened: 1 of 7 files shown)


src/service/AggregateTransactionService.ts
```diff
@@ -24,16 +24,24 @@ import { InnerTransaction } from '../model/transaction/InnerTransaction';
...
  */
 export class AggregateTransactionService {
+    private readonly multisigRepository: MultisigRepository;
+    private readonly networkRepository: NetworkRepository;
     /**
      * Constructor
      * @param multisigRepository
      */
-    constructor(private readonly multisigRepository: MultisigRepository) {}
+    constructor(public readonly repositoryFactory: RepositoryFactory) {
```
fboucquez, line 41: remove public readonly (it just a parm, not a a object field as the fields are the created repositories)

src/service/AggregateTransactionService.ts
```diff
@@ -74,6 +82,41 @@ export class AggregateTransactionService {
...
+    }
+
+    /**
+     * Get next cosignature allowed per aggregate
+     * @returns {Observable<number>}
+     */
+    public getNetworkMaxCosignaturesPerAggregate(): Observable<number> {
+        return this.networkRepository.getNetworkProperties().pipe(
+            map((properties) => {
+                console.log(!properties.plugins.aggregate?.maxCosignaturesPerAggregate);
```
fboucquez, line 111: remove console.log

src/service/AggregateTransactionService.ts
```diff
@@ -74,6 +82,41 @@ export class AggregateTransactionService {
...
+                if (!properties.plugins.aggregate?.maxCosignaturesPerAggregate) {
+                    throw new Error('Cannot get maxCosignaturesPerAggregate from network properties.');
+                }
+                return parseInt(properties.plugins.aggregate?.maxCosignaturesPerAggregate);
```
fboucquez, line 115: A number may be `1'000`, remove possible `'`

http://api-01.us-west-1.symboldev.network:3000/network/properties

rg911: will that ever be over 1000? @dgarcia360

PR #553, opened 2020-05-03
Catapult-rest pagination implemenation

rg911:
- Transaction endpoint only

Feature branch `pagination` created


src/infrastructure/AccountHttp.ts
```diff
@@ -88,136 +83,16 @@ export class AccountHttp extends Http implements AccountRepository {
             dto.account.accountType.valueOf(),
             dto.account.linkedAccountKey,
             dto.account.activityBuckets.map((bucket) => {
-                return new ActivityBucket(bucket.startHeight, bucket.totalFeesPaid, bucket.beneficiaryCount, bucket.rawScore);
+                return new ActivityBucket(
+                    bucket.startHeight,
```
fboucquez, line 87: startHeight must be a UInt64 too

src/infrastructure/Page.ts
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Constructor.
+     *
+     * @param data the page data
+     * @param pageNumber the current page number starting from 1.
+     * @param pageSize the page size.
+     * @param totalEntries the total entries.
+     * @param totalPages the total pages for the given criteria.
+     */
+    constructor(data: T[], pageNumber = 0, pageSize = 0, totalEntries = 0, totalPages = 0) {
```
fboucquez, line 57: I wouldn't put default values to avoid missing one of them

src/infrastructure/TransactionSearchCriteria.ts
```diff
@@ -0,0 +1,217 @@
...
+    public getOffset(): string | undefined {
+        return this.offset;
+    }
+
+    public getGroup(): TransactionGroupSubsetEnum | undefined {
+        return this.group;
+    }
+
+    public getTransactionTypes(): TransactionTypeEnum[] | undefined {
+        return this.transactionTypes?.map((type) => type.valueOf());
```
fboucquez, line 115: Move to the TransactionHttp, model should not know about TransactionTypeEum if possible

src/infrastructure/TransactionSearchCriteria.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    public getRecipientAddress(): string | undefined {
+        return this.recipientAddress?.plain();
+    }
+
+    public getSignerPublicKey(): string | undefined {
+        return this.signerPublicKey;
+    }
+
+    public getHeight(): string | undefined {
```
fboucquez, line 102: Why string?

src/infrastructure/TransactionSearchCriteria.ts
```diff
@@ -0,0 +1,217 @@
...
+import { TransactionType } from '../model/transaction/TransactionType';
+import { TransactionGroupSubsetEnum } from 'symbol-openapi-typescript-node-client/dist/model/transactionGroupSubsetEnum';
+import { TransactionTypeEnum } from 'symbol-openapi-typescript-node-client/dist/model/transactionTypeEnum';
+import { Order } from 'symbol-openapi-typescript-node-client';
+
+/**
+ * Defines the params used to search transactions. With this criteria, you can sort and filter
+ * transactions queries using rest.
+ */
+export class TransactionSearchCriteria extends SearchCriteria {
```
fboucquez, line 29: we are considering moving this object to an interface with just public read-only attributes. The user can create objects using object initialization. Examples:

````
const criteria: TransactionSearchCriteria = { address: account.address, orderBy: Order.Asc };`
````
Builders:

````
const criteria1 = new TransactionSearchCriteria().buildAddress(account.address).buildOrder(Order.Asc);
````

@decentraliser what do you think?

The benefit of the first one is a more "TS" notation, more functional programming tuple style, plus the attributes can be readonly (immutable objects in terms of typescript checks)

File: src/infrastructure/TransactionHttp.ts [hunk 6]
```diff
@@ -208,4 +194,63 @@ export class TransactionHttp extends Http implements TransactionRepository {
...
+     * This method maps a TransactionPage from rest to the SDK's Page model object.
+     *
+     * @internal
+     * @param {TransactionPage} transactionPage the TransactionPage object from rest.
+     * @returns {Page<Transaction>} a Page<T> model
+     */
+    private toPage(transactionPage: TransactionPage): Page<Transaction> {
+        return new Page<Transaction>(
+            transactionPage.data.map((data) => CreateTransactionFromDTO(data)),
+            transactionPage.pagination?.pageNumber,
```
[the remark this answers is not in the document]
rg911: if expectation needs to be raised, I would change the api doc

src/infrastructure/Page.ts
rg911: this needs to be changed on doc first I think as the doc defined as optional.

PR #570, opened 2020-05-18
Listener (WS) refactoring

### LINKED ISSUE #569: Aggregate bonded partialAdd ws listener filter issue
#### Issue: 
- Listener channels currently only filter on 'signer' and 'recipientAddress' in TransferTransaction only.
- Does not resolve alias (unresolvedAddress). e.g. Transfer's recipientAddress is an instance of NamespaceId

rg911:
#### Issue: 
- Listener channels currently only filter on 'signer' and 'recipientAddress' in TransferTransaction only.
- Does not resolve alias (unresolvedAddress). e.g. Transfer's recipientAddress is an instance of NamespaceId

#### Fixed:
- Added abstract method in Transactions for listener notification filtering purposes. 
- Filters on address / publicAccount / publicKeys on all transactions. e.g. 'targetAccount' in MetadataTransactions.
- Added `NamespaceRepository` to `Listener` to resolve alias.
Fixed #569

### DIFF
--- a/src/infrastructure/Listener.ts
+++ b/src/infrastructure/Listener.ts
@@ -14,30 +14,29 @@
  * limitations under the License.
  */
 
-import { Observable, Subject } from 'rxjs';
-import { filter, map, share, withLatestFrom } from 'rxjs/operators';
+import { Observable, of, OperatorFunction, Subject } from 'rxjs';
+import { filter, flatMap, map, share } from 'rxjs/operators';
 import * as WebSocket from 'ws';
 import { Address } from '../model/account/Address';
-import { PublicAccount } from '../model/account/PublicAccount';
 import { BlockInfo } from '../model/blockchain/BlockInfo';
+import { NamespaceName } from '../model/namespace/NamespaceName';
 import { AggregateTransaction } from '../model/transaction/AggregateTransaction';
 import { CosignatureSignedTransaction } from '../model/transaction/CosignatureSignedTransaction';
 import { Deadline } from '../model/transaction/Deadline';
 import { Transaction } from '../model/transaction/Transaction';
 import { TransactionStatusError } from '../model/transaction/TransactionStatusError';
-import { UInt64 } from '../model/UInt64';
+import { BlockHttp } from './BlockHttp';
 import { IListener } from './IListener';
-import { CreateTransactionFromDTO } from './transaction/CreateTransactionFromDTO';
 import { NamespaceRepository } from './NamespaceRepository';
-import { NamespaceName } from '../model/namespace/NamespaceName';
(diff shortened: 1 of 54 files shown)


src/model/transaction/TransferTransaction.ts
```diff
@@ -315,4 +315,19 @@ export class TransferTransaction extends Transaction {
...
     }
+
+    /**
+     * @internal
+     * Check a given address should be notified in websocket channels
+     * @param address address to be notified
+     * @param alias address alias (names)
+     * @returns {boolean}
+     */
+    public NotifyAccount(address: Address, alias: NamespaceId[]): boolean {
```
dgarcia360, line 326: applies to all ``NotifyAccount`` functions:

- If it's tagged as @internal, shouldn't it be a private method?
- I would expect NotifyAccount to be notifyAccount (camelCase)

src/infrastructure/Listener.ts
```diff
@@ -336,18 +316,76 @@ export class Listener implements IListener {
...
+            );
+        };
+    }
+
+    /**
+     * Checks if the transaction is signer by an address.
+     * @param transition the transaction
+     * @param address the address.
+     */
+    private isSigned(transition: Transaction, address: Address): boolean {
```
dgarcia360, line 385: Does it make sense to move this function within the Transaction model?

dgarcia360, line 385: typo in transition --> transaction

dgarcia360:
@rg911 I like the approach to check also for the aliases assigned.to the account. Good idea adding a filter by transaction hash - looks handy to check if a given transaction has been confirmed or not.

src/model/transaction/TransferTransaction.ts
fboucquez: it should be lowe case

src/infrastructure/Listener.ts
fboucquez: Yes, transaction NotifyAccount should also call this method

src/model/transaction/TransferTransaction.ts
rg911: Internal should be fine, Internal won't work on abstract. will update the method name 👍

src/model/transaction/AccountAddressRestrictionTransaction.ts
```diff
@@ -236,9 +236,9 @@ export class AccountAddressRestrictionTransaction extends Transaction {
      * @param alias address alias (names)
      * @returns {boolean}
      */
-    public NotifyAccount(address: Address, alias: NamespaceId[]): boolean {
```
dgarcia360: (typo) shouldNotifiAccount --> shouldNotifYAccount

PR #583, opened 2020-05-24
Added injectable websocket to repotFactor.createListener()

### LINKED ISSUE #581: block meta.totalFee is undefined
Expect new blocks listener not to throw an error
[This throws as](https://github.com/nemtech/symbol-sdk-typescript-javascript/blob/2b78872c5f06cf9a4f8d35ccd9e32d3137c59fd0/src/infrastructure/BlockHttp.ts#L106) `dto.meta.totalFee is undefined`

open a block listener and wait for each new block

3. Specifications like the version of the project, operating system, or hardware.
sdk 0.19.1 on latest testnet 951,

### LINKED ISSUE #582: Add injectable websocket as parameter in RepoFactory.createListener()
https://nem2.slack.com/archives/C9YKR0EUX/p1590254564270600?thread_ts=1590245835.269800&cid=C9YKR0EUX

### LINKED ISSUE #584: stateHashSubCacheMerkleRoots is missing in blockInfo

### LINKED ISSUE #585: Refactor RepositoryFactory parameter
- Move all optional parameters to an interface, keeping only 1 mandatory param (URL)
- Add `websocketUrl` as an optional param for ws URL injection

rg911:
- Listener.newblock is now using `NewBlock` model rather than shared `BlockInfo` with rest payload. So that there will be no `undefined` fields in ws response.
- Moved repositoryFactory constructor optional parameters to a interface
- Added `InjectedWebsocket` and `websocketUrl` parameters to the repositoryFactory constructor as optional parameters
- Changed the `Listener` constructor to use `url` without appending `/ws` suffix.
- Added missing field `stateHashSubCacheMerkleRoots` in `BlockInfo`

Fixed #581
Fixed #582
Fixed #584
Fixed #585


fboucquez:
What about moving the parameters to the repository factory constructor? The clients will only have a centralised place to configure the factory ans will keep using repositoryFactory.createListener() as they used to.

rg911:
I was thinking this is quite specific for listener only, probably better keep in the method instead?

Issue #588, opened 2020-05-26
Revisit how the TS SDK runs in a vanilla JS Browser


fboucquez:
It seems that the user needs to polyfill a client app when testing the SDK. This may be too convoluted and not friendly for quick development.
This story is to revisit how the SDK can run in a browser with just the JS compiled files and how the client can provide some of the dependencies (WebSocket for example) to avoid babel/polyfill 

The outcome of this story should be:
- Reduce the requirements needed to run the SDK
- Document those requirements and how to set up them if necessary.

evias:
the fix **can be** trivial ; using a build:browser npm script we can just add webpack, which will bundle the sdk into a “commonjs” package named SymbolSDK that is then added onto window for browser compatibility. This process is also done in symbol-qr-library, using the npm script `build:browser` you will see it basically just runs `webpack` with a custom configuration.

there is another way to do this, with browserify …. but browserify is less "magic" … and comes with many “polyfill-x” and “polyfill-y” dependencies needs (e.g. `tls-browserify`, `net-browserify`, `level-fs-browser`, etc.)

webpack is just “npm install webpack” and then run webpack with correct configuration.

I will open a proposal PR in next hour to reflect a _proposal_ fix.

PR #589, opened 2020-05-26
Added webpack bundle build and configuration

evias:
Fix for browser compatibility mentioned in #588

Added `webpack` to bundle a browser build of the SDK and map it to `window.SymbolSDK`.

No upload of said bundle have been added or thought through yet.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -546,6 +546,193 @@
         }
       }
     },
+    "@webassemblyjs/ast": {
+      "version": "1.9.0",
+      "resolved": "https://registry.npmjs.org/@webassemblyjs/ast/-/ast-1.9.0.tgz",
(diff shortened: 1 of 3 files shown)


fboucquez:
The built output would be different, specific for a browser but not for a node necessary. How to have one artifact compatible with both env? Maybe 2 deployed artifacts.
I think I would detect exactly the problematic functions/classes and extract them out. Then each client can configure the SDK for the env they are running

evias:
There is also a possibility to use `bundledDependencies` for these cases, if you are going to track down these

Link here: https://docs.npmjs.com/files/package.json#bundleddependencies

fboucquez:
This pr has been superseded by https://github.com/nemtech/symbol-sdk-typescript-javascript/pull/611

@evias If you still have clients that require special webpack or browserify processing please let us now! The plan is for clients to use the SDK as-as for both browser and server.

PR #600, opened 2020-06-02
Pagination for Block, Mosaic and Transaction endpoints

rg911:


src/infrastructure/transaction/CreateTransactionFromDTO.ts
```diff
@@ -122,6 +122,23 @@ const extractMessage = (message: any): PlainMessage | EncryptedMessage => {
...
+ * @param meta - Transaction meta data
+ * @param id - TransactionId
+ * @return {TransactionInfo | AggregateTransactionInfo | undefined}
+ */
+const extractTransactionMeta = (meta: any, id: string): TransactionInfo | AggregateTransactionInfo | undefined => {
+    if (!meta) {
+        return undefined;
+    }
+    if (meta.aggregateHash || meta.aggregateId) {
+        return new AggregateTransactionInfo(UInt64.fromNumericString(meta.height), meta.index, id, meta.aggregateHash, meta.aggregateId);
```
fboucquez, line 138: For the ids, I would do `id || meta.id`. The id can be returned in either the root level or inside the transaction depending on the called endpoint

https://github.com/nemtech/symbol-openapi/issues/145

src/infrastructure/Page.ts
```diff
@@ -0,0 +1,106 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * It represents a page of results after a repository search call.
+ *
+ * @param <T> then model type.
+ */
+export class Page<T> {
```
fboucquez, line 22: I would transform this page to a more TS POJO like the other objects making the fields public read-only. This style is "too Java" and inconsistent with the TS current mode and could have issues if converting it from/to json.

isLast could be an attribute calculated and set in the constructor so it can be serialized/deserialized too.

src/infrastructure/Page.ts
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Constructor.
+     *
+     * @param data the page data
+     * @param pageNumber the current page number starting from 1.
+     * @param pageSize the page size.
+     * @param totalEntries the total entries.
+     * @param totalPages the total pages for the given criteria.
+     */
+    constructor(data: T[], pageNumber = 0, pageSize = 0, totalEntries = 0, totalPages = 0) {
```
fboucquez, line 57: I would remove the defaults, we may miss a require attribute

src/infrastructure/searchCriteria/SearchCriteria.ts
```diff
@@ -0,0 +1,43 @@
...
+     * Filter by page number. (optional)
+     */
+    pageNumber?: number;
+
+    /**
+     * Entry id at which to start pagination. If the ordering parameter is set to -id, the elements
+     * returned precede the identifier. Otherwise, newer elements with respect to the id are
+     * returned. (optional)
+     */
+    offset?: string;
```
fboucquez, line 42: BlockInfo and MosaicInfo are not exposing the database id used to set this offset when pagination from a fix point. The database ids would need to be mapped from the id root attributes in the Block and Mosaic Rest DTO. To avoid confusion with the natural id, I called the `recordId` in java

PR #634, opened 2020-07-20
Task/g628 metadata receipt

rg911:
Applied pagination changes for Metadata and Receipt endpoints


fboucquez:
ReceiptRepository should provide 2 or 3 independent searches so the user knows exactly what are the accepted parameters and the returned types. 

https://github.com/nemtech/symbol-sdk-typescript-javascript/blob/task/g628_metadata_receipt/src/infrastructure/ReceiptRepository.ts#L27

Java as an example:

https://github.com/nemtech/symbol-sdk-java/blob/metadata-receipt/sdk-core/src/main/java/io/nem/symbol/sdk/api/ReceiptRepository.java#L32

Issue #147, opened 2019-05-15
[Dragon 1] features


rg911:
- Generate API and DTO codes using swagger doc from Nem2-doc with latest bugs fix and `Dragon` features.
- Apply to Nem2-lib and test on SDK
  1. Generate latest API client / server codes - Lib
  2. Generate latest api models codes - Lib
  3. Add `recipient` to secretProofTransaction (schema, buffer, transaction) - Lib
  4. Add new http clients (blockHttp, chainHttp, diagnosticHttp) - SDK
  5. Add new models (Server, MerkelPath) - SDK
  6. Add / Migrate previous `Receipts (Dragon)` changes. [PR](https://github.com/nemtech/nem2-sdk-typescript-javascript/pull/137)

- Apply `Dragon` changes
  1. Add `recipient` to secretProofTransaction - SDK
  2. Fix e2e tests (Multisig - need aggregated, SecretProof - new recipient field)
  3. Add e2e tests for new routes (Block, Chain, Diagnostic)
  4. Add / Migrate previous `Receipts (Dragon)` changes. [PR](https://github.com/nemtech/nem2-sdk-typescript-javascript/pull/137)
- Resolve breaking issues.

- Prepend `GenerationHash` to transaction payload before signing

PR #648, opened 2020-08-12
Updated delegated harvesting message payload.

rg911:
- Updated delegated harvesting message marker to `E201735761802AFE`
- Added vrf private key to the message payload, so the encrypted message payload becomes `signingPrivateKey + vrfPrivateKey`

Fixed  - #147

### DIFF
--- a/src/model/message/PersistentHarvestingDelegationMessage.ts
+++ b/src/model/message/PersistentHarvestingDelegationMessage.ts
@@ -31,22 +31,23 @@ export class PersistentHarvestingDelegationMessage extends Message {
     }
 
     /**
-     *
-     * @param delegatedPrivateKey - Private key of delegated account
-     * @param recipientPublicKey - Recipient public key
+     * @param signingPrivateKey - Remote harvester signing private key linked to the main account
+     * @param vrfPrivateKey - VRF private key linked to the main account
+     * @param nodePublicKey - Node certificate public key
      * @param {NetworkType} networkType - Catapult network type
      * @return {PersistentHarvestingDelegationMessage}
      */
     public static create(
-        delegatedPrivateKey: string,
-        recipientPublicKey: string,
+        signingPrivateKey: string,
+        vrfPrivateKey: string,
+        nodePublicKey: string,
         networkType: NetworkType,
     ): PersistentHarvestingDelegationMessage {
         const ephemeralKeypair = Account.generateNewAccount(networkType);
         const encrypted =
             MessageMarker.PersistentDelegationUnlock +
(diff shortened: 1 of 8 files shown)


src/model/message/PersistentHarvestingDelegationMessage.ts
```diff
@@ -62,7 +63,7 @@ export class PersistentHarvestingDelegationMessage extends Message {
     /**
      *
      * @param encryptMessage - Encrypted message to be decrypted
-     * @param privateKey - Recipient private key
+     * @param privateKey - Node certificate private key
      * @return {string}
      */
     public static decrypt(encryptMessage: PersistentHarvestingDelegationMessage, privateKey: string): string {
```
fboucquez, line 69: This method should do the split and return a pair of signingPrivateKey and vrfPrivateKey ready to be consumed by the user (sender sends 2 keys, receipient decode 2 keys).

test/model/message/PersistentHarvestingDelegationMessage.spec.ts
```diff
@@ -89,16 +93,17 @@ describe('PersistentHarvestingDelegationMessage', () => {
             signedTransaction.payload.substring(322, signedTransaction.payload.length),
         );
         const plainMessage = PersistentHarvestingDelegationMessage.decrypt(encryptMessage, recipient.privateKey);
-        expect(plainMessage).to.be.equal(delegatedPrivateKey);
+        expect(plainMessage).to.be.equal(signingPrivateKey + vrfPrivateKey);
```
fboucquez, line 96: plainMessage should be a pair of the 2 private keys already split

src/model/message/PersistentHarvestingDelegationMessage.ts
rg911: I was thinking to remove it since normal user cannot decrypt message since they need the node private key. but it can be used by node owner

test/model/message/PersistentHarvestingDelegationMessage.spec.ts
rg911: yeah. not expecting normal user to decrypt the message, only the node owner, I guess the combined key string should be okay

fboucquez: Fair enough

src/model/message/PersistentHarvestingDelegationMessage.ts
fboucquez: I'll keep it. It's handy for at least unit testing

PR #685, opened 2020-10-03
Fixed #684

### LINKED ISSUE #684: Revisit deprecated and warnings in dependencies
This is the report of install the latest TS SDK:

```
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated request-promise-native@1.0.9: request-promise-native has been deprecated because it extends the now deprecated request package, see https://github.com/request/request/issues/3142
npm WARN deprecated js-joda@1.11.0: This package is deprecated, please check/use @js-joda/core instead
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN symbol-sdk@0.21.0 requires a peer of js-joda@^1.6.2 but none is installed. You must install peer dependencies yourself.
npm WARN symbol-sdk@0.21.0 requires a peer of utf8@^2.1.2 but none is installed. You must install peer dependencies yourself.
npm WARN symbol-sdk@0.21.0 requires a peer of rxjs@^6.5.3 but none is installed. You must install peer dependencies yourself.

```

This task is to revisit and update the dependencies.

rg911:
Updated packages.json


fboucquez:
Do we need the "request" packages @rg911?   Should fetch be enough? What happens if we remove request?

rg911:
> Do we need the "request" packages @rg911? Should fetch be enough? What happens if we remove request?
cannot remember why we used it in the first place. but you are probably right here. Let me have a look

fboucquez:
Probably leftover from the last request implementation we replaced with fetch. Maybe double-check if there is any other not used dep we can remove?

fboucquez (approved):

PR #686, opened 2020-10-08
 removed hardcoded nemesis epoch

### LINKED ISSUE #681: Nemesis epoch time is hard coded in the SDK but can change on the server side.
For the mainnet release, the epochAdjustment value will be change.  This will require an SDK update.  Instead of hardcoded the SDK should read the value from the network config when available. 

[network]
epochAdjustment = 1573430400s

https://github.com/nemtech/symbol-sdk-typescript-javascript/blob/2b18ca2f991aa692dd4384e93b7377954faed695/src/model/transaction/Deadline.ts#L28

rg911:
- removed hardcoded nemesis epoch
- Changed `Deadline.value (LocalDateTime) ` to `Deadline.adjustedValue (number / seconds)`
- Deadline.create(...) now requires `EpochAdjustment (js-joda/Duration`
- Added toLocalDatetime (with custom zone support) in Deadline class
- Added EpochAdjustment getter in RepositoryFactory

Fixed #681

### DIFF
--- a/src/infrastructure/RepositoryFactoryHttp.ts
+++ b/src/infrastructure/RepositoryFactoryHttp.ts
@@ -53,6 +53,7 @@ import { HashLockRepository } from './HashLockRepository';
 import { SecretLockRepository } from './SecretLockRepository';
 import { SecretLockHttp } from './SecretLockHttp';
 import { HashLockHttp } from './HashLockHttp';
+import { Duration } from 'js-joda';
 /**
  * Receipt http repository.
  *
@@ -64,7 +65,7 @@ export class RepositoryFactoryHttp implements RepositoryFactory {
     private readonly websocketUrl: string;
     private readonly websocketInjected?: any;
     private readonly fetchApi?: any;
-    private readonly epochAdjustment: Observable<number>;
+    private readonly epochAdjustment: Observable<Duration>;
 
     /**
      * Constructor
@@ -77,12 +78,12 @@ export class RepositoryFactoryHttp implements RepositoryFactory {
         this.fetchApi = configs?.fetchApi;
         this.networkType = configs?.networkType ? observableOf(configs.networkType) : networkRepo.getNetworkType().pipe(shareReplay(1));
         this.epochAdjustment = configs?.epochAdjustment
-            ? observableOf(configs.epochAdjustment)
+            ? observableOf(Duration.ofSeconds(configs.epochAdjustment))
             : networkRepo
                   .getNetworkProperties()
                   .pipe(
                       map((property) => {
-                          return parseInt(property.network.epochAdjustment?.replace('s', '') ?? '0');
+                          return Duration.ofSeconds(parseInt(property.network.epochAdjustment?.replace('s', '') ?? '0'));
                       }),
                   )
                   .pipe(shareReplay(1));
@@ -145,7 +146,7 @@ export class RepositoryFactoryHttp implements RepositoryFactory {
     }
 
     createTransactionRepository(): TransactionRepository {
-        return new TransactionHttp(this.url, this.epochAdjustment, this.fetchApi);
+        return new TransactionHttp(this.url, this.fetchApi);
     }
 
     createTransactionStatusRepository(): TransactionStatusRepository {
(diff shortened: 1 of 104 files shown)


src/model/transaction/Deadline.ts
```diff
@@ -25,7 +25,7 @@ export class Deadline {
     /**
      * @type {number}
      */
-    public static timestampNemesisBlock = 1573430400;
+    private nemesisEpoch: number;
```
fboucquez, line 28: Hi Steve, I've swapped the internal fields of the Deadline object in java. 

Instead of keeping both values (local and epoch) you can store the server value as-is.  When serializing and deserializing from and back dto and catbuffer you don't need to provide the epochAdjustment (the mappers won't need access to a configuration).  

epochAdjustment is only necessary when users creates a relative deadline from current timestamp (like now + 2 hours).

https://github.com/nemtech/symbol-sdk-java/blob/351_configurable_epoch_time/sdk-core/src/main/java/io/nem/symbol/sdk/model/transaction/Deadline.java#L35

This change would not require a client change.

fboucquez, line 28: I would use the server name of `epochAdjustment` instead of `nemesisEpoch`

src/infrastructure/RepositoryFactoryHttp.ts
```diff
@@ -161,4 +171,8 @@ export class RepositoryFactoryHttp implements RepositoryFactory {
     createListener(): IListener {
         return new Listener(this.websocketUrl, this.createNamespaceRepository(), this.websocketInjected);
     }
+
+    getNemesisEpoch(): Observable<number> {
+        return this.nemesisEpoch;
```
fboucquez, line 176: I would use `epochAdjustment` anme

src/infrastructure/RepositoryFactoryHttp.ts
```diff
@@ -64,18 +64,28 @@ export class RepositoryFactoryHttp implements RepositoryFactory {
...
-            ? observableOf(configs.networkType)
-            : this.createNetworkRepository().getNetworkType().pipe(shareReplay(1));
+        this.networkType = configs?.networkType ? observableOf(configs.networkType) : networkRepo.getNetworkType().pipe(shareReplay(1));
+        this.nemesisEpoch = configs?.nemesisEpoch
+            ? observableOf(configs.nemesisEpoch)
+            : networkRepo
+                  .getNetworkProperties()
+                  .pipe(
+                      map((property) => {
+                          return parseInt(property.network.epochAdjustment?.replace('s', '') ?? '0');
```
fboucquez, line 85: Server properties may be defined in ms, seconds, hours, days, etc. Soner rather than later we would need to support multiple duration time. 

This is my implementation in java

https://github.com/nemtech/symbol-sdk-java/blob/351_configurable_epoch_time/sdk-core/src/main/java/io/nem/symbol/core/utils/FormatUtils.java#L95

src/infrastructure/transaction/CreateTransactionFromDTO.ts
```diff
@@ -439,7 +441,7 @@ const CreateStandaloneTransactionFromDTO = (transactionDTO, transactionInfo): Tr
         return new MosaicMetadataTransaction(
             transactionDTO.network,
             transactionDTO.version,
-            extractDeadline(transactionDTO.deadline),
+            extractDeadline(nemesisEpoch, transactionDTO.deadline),
```
fboucquez, line 444: If deadline stores deadline from server as-as, you don't need to provide the nemesisEpech when mapping

fboucquez (changes requested):

src/model/transaction/Deadline.ts
fboucquez: `public value: LocalDateTime;` would not apply anymore. 

we could have a 

public toLocalDateTime(epochAdjustment:number):LocalDateTime that converts the as-is server value back to a real "date"

src/infrastructure/RepositoryFactoryHttp.ts
rg911: Where does this come from? I don't think this will be changed anymore since testnet is already launched.

fboucquez (approved):

PR #687, opened 2020-10-09
Added merkle hash calculation for block transactions

### LINKED ISSUE #620: BLS Voting key generation

rg911:
- Fixed #620

### DIFF
--- a/src/service/BlockService.ts
+++ b/src/service/BlockService.ts
@@ -115,7 +115,6 @@ export class BlockService implements IBlockService {
      * @returns calculated root hash
      */
     private getTransacactionMerkleRoot(transactions: Transaction[]): string {
-        console.log(transactions);
         const leaves = transactions
             .sort((n1, n2) => n1.transactionInfo!.index - n2.transactionInfo!.index)
             .map((transaction) => transaction.transactionInfo!.hash);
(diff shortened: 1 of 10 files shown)


src/service/BlockService.ts
```diff
@@ -87,4 +107,27 @@ export class BlockService implements IBlockService {
...
     }
+
+    /**
+     * @internal
+     * Create merkle root hash for block transactions
+     * @param transactions Block transactions
+     * @returns calculated root hash
+     */
+    private getTransacactionMerkleRoot(transactions: Transaction[]): string {
+        console.log(transactions);
```
fboucquez, line 118: Remove console.log

src/service/BlockService.ts
```diff
@@ -64,6 +71,19 @@ export class BlockService implements IBlockService {
...
+    /**
+     * Calculate transactions merkle root hash from a block
+     * @param height block height
+     * @returns root hash
+     */
+    public calculateTransactionsMerkleRootHash(height: UInt64): Observable<string> {
+        const streamer = new TransactionPaginationStreamer(this.transactionRepository);
+        return streamer
+            .search({ group: TransactionGroup.Confirmed, height: height })
+            .pipe(toArray())
```
fboucquez, line 83: If possible, avoid toArray(). This puts all the transactions of the given height in memory. Can rest sort by index?

If not possible (due to the sort below), map them to ({index, hash}) before doing the to Array. Then sort that list of pair object

fboucquez (changes requested):
Also add an e2e test validating that server resolves the same hash. I'll copy the implementation over to java :)

fboucquez (approved):
Great!

PR #695, opened 2020-10-19
Currency and CurrencyService

fboucquez:
Hi Steve, I've done a small refactoring to the NetworkCurrency object to align with Java.

NetworkCurrency is an object that knows how to create Mosaic (it's not a subclass of mosaic anymore). Now we can create NetworkCurrency based on the network configuration. 

I've removed cat.currency and cat.harvest from the production code as these names are just generic that can be easily changed via bootstrap. 

There is a NetworkCurrencyService that knows how to load/create NetworkCurrency objects. NetworkCurrency could also be loaded from user-provided mosaic id if desired. 

Fixes https://github.com/nemtech/symbol-sdk-typescript-javascript/issues/692
Fixes https://github.com/nemtech/symbol-sdk-typescript-javascript/issues/693


src/service/interfaces/INetworkCurrencyService.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Observable } from 'rxjs';
+import { MosaicId } from '../../model/mosaic';
+import { NetworkCurrency } from '../../model/mosaic';
+import { NetworkCurrencies } from '../../model/mosaic/NetworkCurrencies';
+
+export interface INetworkCurrencyService {
+    /**
+     *
+     */
+    getMainNetworkCurrencies(): Observable<NetworkCurrencies>;
```
rg911, line 25: `getPublicNetworkCurrency` would be better

src/service/interfaces/INetworkCurrencyService.ts
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+import { Observable } from 'rxjs';
+import { MosaicId } from '../../model/mosaic';
+import { NetworkCurrency } from '../../model/mosaic';
+import { NetworkCurrencies } from '../../model/mosaic/NetworkCurrencies';
+
+export interface INetworkCurrencyService {
+    /**
+     *
```
rg911, line 23: comments missing?

src/service/interfaces/INetworkCurrencyService.ts
```diff
@@ -0,0 +1,28 @@
...
+     */
+    getMainNetworkCurrencies(): Observable<NetworkCurrencies>;
+
+    loadCurrencies(mosaicIds: MosaicId[]): Observable<NetworkCurrency[]>;
```
rg911, line 27: should it use `get...` for consistency?
also missing docs

rg911 (changes requested):
Looks very good. 
commented on some naming conventions and docs missing

fboucquez:
@rg911 I've renamed `NetworkCurrency` with `Currency`. `Currency` is a generic object that creates mosaics. They may be a network mosaic/alias or a custom/user mosaic/alias. `NetworkCurrencies` is still there because it holds the network currencies main and harvest.

I've also done the fixes as requested

fboucquez:
@rg911 could you review? This PR doesn't require a rest update like the next ones.

src/model/mosaic/Currency.ts
```diff
@@ -0,0 +1,129 @@
...
+import { NamespaceId } from '../namespace/NamespaceId';
+import { UInt64 } from '../UInt64';
+import { Mosaic } from './Mosaic';
+import { MosaicId } from './MosaicId';
+import { UnresolvedMosaicId } from './UnresolvedMosaicId';
+
+/**
+ * An object that knows how to create Mosaics based on the Mosaic Info and Namespace configuration.
+ */
+export class Currency {
```
rg911, line 26: Shall we call it NetworkCurrency instead of Currency? to differentiate with other tokens might be created by users

fboucquez: That's the thing, Currency supports user-created mosaics...

rg911 (approved):

PR #700, opened 2020-10-26
Fixed alias resolution in Namespace Repository

fboucquez:
Fixes https://github.com/nemtech/symbol-sdk-typescript-javascript/issues/699

### DIFF
--- a/src/infrastructure/NamespaceHttp.ts
+++ b/src/infrastructure/NamespaceHttp.ts
@@ -13,10 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Observable } from 'rxjs';
+import { Observable, of, throwError } from 'rxjs';
 import { mergeMap } from 'rxjs/operators';
 import { NamespaceDTO, NamespaceInfoDTO, NamespaceRoutesApi } from 'symbol-openapi-typescript-fetch-client';
-import { Convert as convert, RawAddress as AddressLibrary } from '../core/format';
 import { AccountNames } from '../model/account/AccountNames';
 import { Address } from '../model/account/Address';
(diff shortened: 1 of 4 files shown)


src/infrastructure/NamespaceHttp.ts
```diff
@@ -168,24 +167,18 @@ export class NamespaceHttp extends Http implements NamespaceRepository {
...
-        return this.networkTypeObservable.pipe(
-            mergeMap(() =>
-                this.call(this.namespaceRoutesApi.getNamespace(namespaceId.toHex()), (body) => {
-                    const namespaceInfoDTO = body;
-                    if (namespaceInfoDTO.namespace === undefined) {
-                        // forward catapult-rest error
-                        throw namespaceInfoDTO;
+        return this.getNamespace(namespaceId).pipe(
+            mergeMap((info) => {
+                if (info.depth === 1) {
```
rg911, line 172: Doesn't look quite right to me here. Why check level1? and also hardcoded to 1?

rg911 (changes requested):

src/infrastructure/NamespaceHttp.ts
fboucquez: As far as I can see, if the depth is > 1, the info is an alias of an alias (AB2F0159E229A0BD), and the alias type is 0

http://api-01.ap-southeast-1.0.10.0.x.symboldev.network:3000/namespaces/AB2F0159E229A0BD

If you go to the level0 (root alias E66AE40356B62BFC) you can retrieve the information of the final alias and get the real mosaic id or address

http://api-01.ap-southeast-1.0.10.0.x.symboldev.network:3000/namespaces/E66AE40356B62BFC

fboucquez:
Invalid resolution

PR #731, opened 2020-11-30
All transaction related notification should take an unresolved address

james-lin71:
Resolve transaction notifications with unresolved address.


src/infrastructure/IListener.ts
```diff
@@ -121,26 +121,26 @@ export interface IListener {
      * @param transactionHash the transaction hash filter.
```
fboucquez, line 121: invalid address name in docs

fboucquez (changes requested):
Good job @james-lin71 . Is it possible to add unit tests to Listener.spec.ts?

fboucquez (approved):

PR #752, opened 2021-01-25
Notification multisig

### LINKED ISSUE #751: Add integrated service for multisig + cosigner ws listener

rg911:
- Subscribe to multisig account from cosigner 
- Fixed #751

### DIFF
--- a/src/infrastructure/Listener.ts
+++ b/src/infrastructure/Listener.ts
@@ -31,6 +31,8 @@ import { Transaction } from '../model/transaction/Transaction';
 import { TransactionStatusError } from '../model/transaction/TransactionStatusError';
 import { UInt64 } from '../model/UInt64';
 import { IListener } from './IListener';
+import { MultisigHttp } from './MultisigHttp';
+import { MultisigRepository } from './MultisigRepository';
 import { NamespaceRepository } from './NamespaceRepository';
 import { CreateTransactionFromDTO } from './transaction/CreateTransactionFromDTO';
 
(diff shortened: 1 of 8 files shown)


src/infrastructure/Listener.ts
```diff
@@ -92,9 +94,14 @@ export class Listener implements IListener {
...
          */
         private websocketInjected?: any,
+        /**
+         * Multisig repository for resolving multisig accounts
+         */
+        private multisigRepository?: MultisigRepository,
     ) {
         this.url = url.replace(/\/$/, '');
         this.messageSubject = new Subject();
+        this.multisigRepository = this.multisigRepository ? this.multisigRepository : new MultisigHttp(this.namespaceRepository.getUrl());
```
fboucquez, line 104: Why asking for the url to the node repository? The URL is up there plus I would like to keep the repository URL agnostic.

fboucquez: It's kind of hacky. I would make multisigRepository required like namespaceRepository and remove getUrl from the namespace repository. If somebody is already using the ListnerHttp, this person would know how to create a multisigRepository.

fboucquez:
@rg911 , there is channel named 'modifyMultisigAccount'. But unused? are we missing listeners?

src/infrastructure/Listener.ts
rg911: is the way not to introduce a breaking change at this point of time

fboucquez: getUrl is breaking the abstraction. Also why in the NamespaceRepository and not in the other ones (I'm not saying they should have it, it's just inconsistent). 

If the devs coded the clients correctly, the construction of the Listener should be behind the scenes without breaking the code. If not, most people would know how to create the repositories used by the listener.

Those are my points, if you like to merge it's all good.

PR #761, opened 2021-02-09
Filter ws transaction message by unresolved address

### LINKED ISSUE #760: Use UnresolvedAddress in websocket listener
Due to catapult-rest patch the ws subscriber, the sdk should use UnresolvedAddress for ws message filtering.

rg911:
- Removed NamespaceName api call
- filter transaction by unresolved address
- - Fixed #760

### DIFF
--- a/src/infrastructure/Listener.ts
+++ b/src/infrastructure/Listener.ts
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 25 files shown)


fboucquez (approved):
Approved after a chat with Steve. It doesn't affect the behavior of the listener too much. 

The Java listener could be used for reference if we need to change the behavior

https://github.com/nemtech/symbol-sdk-java/pull/389

PR #766, opened 2021-02-17
Add node version model

OlegMakarenko:
Resolved: #765

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "symbol-sdk",
-    "version": "0.23.2",
+    "version": "0.23.3",
     "lockfileVersion": 1,
     "requires": true,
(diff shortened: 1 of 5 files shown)


rg911 (commented):
What is the main purpose of this class? Convert/formatting or used as a data / dto model ? 
If this is only used for conversion, could it be created inside the `utility` or ` formatter` instead and make it more generic?

fboucquez (approved):
The class is to solve this issue https://github.com/nemgrouplimited/symbol-node-rewards/issues/162#issuecomment-779994288

It could be in a helper class but I think it's pretty nice. It follows the Address.ts and MosaicId.ts classes. Docs and Unit tests are awesome.

rg911 (approved):

PR #782, opened 2021-05-19
fix: UnhandledPromiseRejection errors

fboucquez:
Exception handling simplification in http calls
Defer when loading node info metadata
--unhandled-rejections=strict in tests

Fixes https://github.com/nemtech/symbol-sdk-typescript-javascript/issues/781


test/infrastructure/RepositoryFactory.spec.ts
```diff
@@ -70,6 +72,17 @@ describe('RepositoryFactory', () => {
         expect(repositoryFactory.createFinalizationRepository()).to.be.not.null;
     });
 
+    it('Raise error without unhandled-rejections', async () => {
+        const url = 'https://www.google.com';
+        const repositoryFactory = new RepositoryFactoryHttp(url);
+        try {
+            await repositoryFactory.createNodeRepository().getNodeHealth().toPromise();
+            expect(true).to.be.false;
```
yilmazbahadir, line 80: Should we use [chai-as-promised](https://www.chaijs.com/plugins/chai-as-promised/) or a similar promise assertion library as a better practice in general for promise assertions?

test/infrastructure/RepositoryFactory.spec.ts
```diff
@@ -70,6 +72,17 @@ describe('RepositoryFactory', () => {
         expect(repositoryFactory.createFinalizationRepository()).to.be.not.null;
     });
 
+    it('Raise error without unhandled-rejections', async () => {
+        const url = 'https://www.google.com';
```
yilmazbahadir, line 76: Any other way to achieve this without relying on a remote call? I think we can use `sinon` to stub the `getNodeHealth`method to throw an error.

test/infrastructure/RepositoryFactory.spec.ts
fboucquez: Thanks! chai as promises looks nice. Although the try-catch could be more end-user-style assertion? And I wouldn't add another lib just yet. Let's consider it for the future.

test/infrastructure/RepositoryFactory.spec.ts
fboucquez: TS SDK uses ts-mockito. I've mocked the low-level generated NodeRoutesApi reflecting how it rejects responses with the current mock lib. The (nodeRepository as any).nodeRoutesApi doesn't look nice, but well.

rg911 (approved):

yilmazbahadir (approved):

PR #788, opened 2021-05-28
feat: transaction signWith method broken into smaller  methods

yilmazbahadir:
This PR is a proposal of breaking down the` Transaction.signWith` method into smaller methods(corresponding to the steps)

There are three parts of signing:
1- Preparing the byte array
2- Signing the byte array with private key (this happens on hardware wallet in some cases)
3- Preparing the signed raw tx - payload 
4- Return SignedTransaction object

### DIFF
--- a/src/model/transaction/Transaction.ts
+++ b/src/model/transaction/Transaction.ts
@@ -214,16 +214,12 @@ export abstract class Transaction {
     public signWith(account: Account, generationHash: string): SignedTransaction {
         const generationHashBytes = Array.from(Convert.hexToUint8(generationHash));
         const byteBuffer = Array.from(this.generateBytes());
+        // 1. prepare the raw tx to be signed
         const signingBytes = this.getSigningBytes(byteBuffer, generationHashBytes);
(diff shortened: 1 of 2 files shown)


src/model/transaction/Transaction.ts
```diff
@@ -233,6 +229,34 @@ export abstract class Transaction {
         );
     }
 
+    /**
+     * Signs raw transaction with the given private key
+     * @param {string} privateKey - Private key of the signer account
+     * @param {number[]} rawTxSigningBytes - Raw transaction siging bytes
+     * @returns {number[]} Signature byte array
+     */
+    public static signRawTx(privateKey: string, rawTxSigningBytes: number[]): number[] {
```
fboucquez, line 238: Should the public method use Uint8Array for input and output rather than number[]?

yilmazbahadir: Makes sense, updated.

fboucquez (approved):

rg911 (approved):

PR #800, opened 2021-07-17
Multisig multilevel subscription

### LINKED ISSUE #798: Multisig multi-level subscription
Atm SDK listeners are not handling multilevel multisig which limits the listener to watch the first level rather than the whole 3 levels if they do exist.

https://github.com/nemtech/symbol-sdk-typescript-javascript/blob/1bb1d1c220ade1d11fff2c78e87fec5da55e5a79/src/infrastructure/Listener.ts#L599

Example:
Create accounts A,B,C,D,E
- A and B are cosigners of C
- C and E is a cosigner D

initiate aggregate bonded from E as a cosigner of D.

expected behavior:
- A and B already subscribed, should get notified and be able to sign.

current behavior:
- A and B not subscribed to D so they don't get notified about children's transactions.

bassemmagdy:
- fix #798.
- update rxjs version to be compatible with wallet.


src/core/utils/MultisigGraphUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+import { MultisigAccountGraphInfo } from '../../model/account/MultisigAccountGraphInfo';
+import { MultisigAccountInfo } from '../../model/account/MultisigAccountInfo';
+
+/**
+ * Hash utilities for SecretLock hashing
+ */
+export class MultisigGraphUtils {
+    public static parseObjectProperties(obj, parse): any {
```
rg911, line 24: - missing docs
- `any` type?

what does this method do?

src/core/utils/MultisigGraphUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+            if (typeof obj[k] === 'object' && obj[k] !== null) {
+                this.parseObjectProperties(obj[k], parse);
+            } else if (Object.prototype.hasOwnProperty.call(obj, k)) {
+                parse(k, obj[k]);
+            }
+        }
+    }
+    public static getMultisigChildren(
+        multisigAccountGraphInfoMapped: MultisigAccountInfo[][],
+    ): Array<{ address: string; children: string[] }> {
```
rg911, line 35: why array<T>, can a model class/interface be used here?

src/core/utils/MultisigGraphUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MultisigAccountGraphInfo } from '../../model/account/MultisigAccountGraphInfo';
+import { MultisigAccountInfo } from '../../model/account/MultisigAccountInfo';
+
+/**
+ * Hash utilities for SecretLock hashing
```
rg911, line 21: ??

rg911 (changes requested):

src/core/utils/MultisigGraphUtils.ts
bassemmagdy: @rg11 I didn't think that we need to use a new model/type to handle such a case, do we?

src/core/utils/MultisigGraphUtils.ts
bassemmagdy: It simply helps traverse the tree object and for the return type, changed it to void.

src/core/utils/MultisigGraphUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import { MultisigAccountGraphInfo } from '../../model/account/MultisigAccountGraphInfo';
+import { MultisigAccountInfo } from '../../model/account/MultisigAccountInfo';
+
+/**
+ * MultisigGraph utilities
+ */
+export class MultisigGraphUtils {
+    // Traverses the tree object to pick addresses strings
+    public static parseObjectProperties(obj, parse): void {
```
rg911, line 25: this method looks like generic util rather than specifically for the multisig, The multisig graph has it's own class and properties, is it not possible to parse from it? Otherwise, should this be placed in the global util class / repo instead?

src/core/utils/MultisigGraphUtils.ts
rg911: `Array<{ address: string; children: string[] }>` is type friendly, probably fit the wallet's purpose, but would be nice to create a type / interface for it

src/core/utils/MultisigGraphUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+ * limitations under the License.
+ */
+
+import { MultisigAccountGraphInfo } from '../../model/account/MultisigAccountGraphInfo';
+import { MultisigAccountInfo } from '../../model/account/MultisigAccountInfo';
+
+/**
+ * MultisigGraph utilities
+ */
+export class MultisigGraphUtils {
```
rg911, line 23: None of the methods were properly documented, please consider adding docs (follow js/ts -doc pattern)

segfaultxavi: Yes yes yes yes yes please

src/core/utils/MultisigGraphUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+// Type for Multisig Tree children Object
+export type MultisigChildrenTreeObject = {
+    address: string;
+    children: []; // children array.
+};
+
+export class MultisigGraphUtils {
+    /**
+     * creates a structred Tree object containing Current multisig account with children
+     *@param {MultisigAccountInfo[][]} multisigEnteries
```
rg911, line 33: missing space

rg911: and same as below

rg911 (changes requested):

src/core/utils/MultisigGraphUtils.ts
bassemmagdy: updated.

rg911 (approved):

PR #804, opened 2021-07-26
Allow rest base 32 addresses

fboucquez:
This is the SDK side of https://github.com/symbol/catapult-rest/pull/611

The new rest mappers will accept both encoded/hex and plain/decoded addresses. It will buffer the migration allowing the SDK rest client to be compatible with old and new rests. Eventually, we should only allow base32 addresses. Apps only need to upgrade the SDK, no code change is required.

A migration plan could be:
- Merge and release this SDK fix.
- Merge and release the Rest fix.
- Upgrade SDKs in all our apps (CLIs, explorers, wallets, etc). Release the apps. The apps will be old and new rest compatible. 
- Migrate testnet, mainnet, and community nodes to the new rest when possible.  

No flag is required. 

Related issues:
https://github.com/symbol/symbol-sdk-typescript-javascript/issues/801
https://github.com/symbol/symbol-sdk-typescript-javascript/issues/802

### DIFF
--- a/src/core/utils/DtoMapping.ts
+++ b/src/core/utils/DtoMapping.ts
@@ -34,6 +34,7 @@ import { MerkleTreeBranch } from '../../model/state/MerkleTreeBranch';
 import { MerkleTreeBranchLink } from '../../model/state/MerkleTreeBranchLink';
 import { MerkleTreeLeaf } from '../../model/state/MerkleTreeLeaf';
 import { MerkleTreeNodeType } from '../../model/state/MerkleTreeNodeType';
+import { Convert } from '../format';
 
 export class DtoMapping {
     /**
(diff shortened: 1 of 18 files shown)


src/core/utils/DtoMapping.ts
```diff
@@ -76,6 +77,22 @@ export class DtoMapping {
...
+     *
+     * This method handles both format, encoded (deprecated) and base32 encoded addresses.
+     *
+     * Clients using this SDK will be able to process both payloads.
+     *
+     * @param value the address in encoded (6823BB7C3C089D996585466380EDBDC19D4959184893E38C) format or decoded/plain format (SB3KUBHATFCPV7UZQLWAQ2EUR6SIHBSBEOEDDDF3)
+     * @return the Address object.
+     */
+    public static toAddress(value: string): Address {
+        return Convert.isHexString(value, 48) ? Address.createFromEncoded(value) : Address.createFromRawAddress(value);
```
AnthonyLaw, line 93: If the recipientAddress is alias `THBIMC3THGH5RUYAAAAAAAAAAAAAAAAAAAAAAAA`

It will throw an error `Address Network unsupported` right?

fboucquez: We haven't removed private/mijin network support from the old SDKs yet. Unsure if it makes sense to remove them, this SDK would fade off eventually.

AnthonyLaw: Ignore my question. 
Actually, I misunderstood, even though the alias namespace is still converted to a valid address format from REST.

AnthonyLaw (approved):

PR #822, opened 2021-11-11
fix: Upgraded catbuffer to the latest schemas names

fboucquez:
fix: Upgraded catbuffer to generated classes using the latest schema names

The latest catbuffer schema and parsers changed a few names, mostly from Key -> PublicKey. This PR reflects the changes required to use new generated classes.

The new catbuffer lib is the one generated in https://github.com/symbol/catbuffer-generators/pull/166

Note: Revokable mosaic support would be added in another PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     },
     "dependencies": {
         "@js-joda/core": "^3.2.0",
-        "catbuffer-typescript": "1.0.0",
+        "catbuffer-typescript": "^1.0.1-alpha-202111091444",
         "crypto-js": "^4.0.0",
         "futoin-hkdf": "^1.3.2",
         "js-sha256": "^0.9.0",
--- a/src/model/account/PublicAccount.ts
+++ b/src/model/account/PublicAccount.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { KeyDto } from 'catbuffer-typescript';
+import { PublicKeyDto } from 'catbuffer-typescript';
 import { KeyPair } from '../../core/crypto';
 import { Convert } from '../../core/format';
 import { NetworkType } from '../network/NetworkType';
@@ -101,7 +101,7 @@ export class PublicAccount {
     /**
      * Create Builder object
      */
-    toBuilder(): KeyDto {
-        return new KeyDto(Convert.hexToUint8(this.publicKey));
+    toBuilder(): PublicKeyDto {
+        return new PublicKeyDto(Convert.hexToUint8(this.publicKey));
     }
 }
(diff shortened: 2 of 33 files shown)


rg911 (approved):

rg911 (changes requested):
Check out the travis log, errors in the transaction helper.

fboucquez:
> Check out the travis log, errors in the transaction helper.

fixed package-lock

src/model/account/PublicAccount.ts
```diff
@@ -101,7 +101,7 @@ export class PublicAccount {
     /**
      * Create Builder object
      */
-    toBuilder(): KeyDto {
-        return new KeyDto(Convert.hexToUint8(this.publicKey));
+    toBuilder(): PublicKeyDto {
```
Jaguar0625, line 104: isn't this a breaking change to clients?

package.json
```diff
@@ -94,7 +94,7 @@
     },
     "dependencies": {
         "@js-joda/core": "^3.2.0",
-        "catbuffer-typescript": "1.0.0",
+        "catbuffer-typescript": "^1.0.1-alpha-202111091444",
```
Jaguar0625, line 97: why are we depending on alpha libs?

src/model/transaction/NamespaceRegistrationTransaction.ts
```diff
@@ -217,7 +217,7 @@ export class NamespaceRegistrationTransaction extends Transaction {
     protected createBuilder(): TransactionBuilder {
         let transactionBuilder: NamespaceRegistrationTransactionBuilder;
         if (this.registrationType === NamespaceRegistrationType.RootNamespace) {
-            transactionBuilder = NamespaceRegistrationTransactionBuilder.createNamespaceRegistrationTransactionBuilderRoot(
+            transactionBuilder = NamespaceRegistrationTransactionBuilder.createNamespaceRegistrationTransactionBuilderROOT(
```
Jaguar0625, line 220: do you intend to have ROOT and CHILD capitalized in function names?

package.json
fboucquez: This PR goes together with https://github.com/symbol/catbuffer-generators/pull/166. Once that pr (and the revokable mosaic support) is merged we can release it and upgrade the versions here.

src/model/account/PublicAccount.ts
fboucquez: Yes, those methods should have been `private / @internal` in this SDK I guess.  It's a secondary effect of the schema name changes.

src/model/transaction/NamespaceRegistrationTransaction.ts
fboucquez: I could change the catbuffer generator to lower case those constants like it used to be before the schema renaming. Those methods are internal in this SDK though.

src/model/account/PublicAccount.ts
Jaguar0625: so, expectation is that no one should be calling this publicly?

Jaguar0625 (approved):

rg911 (approved):

src/model/account/PublicAccount.ts
fboucquez: I would be really really surprised.

PR #833, opened 2022-02-06
fix: metadata value non-ascii utf8 encoding issue fixed

### LINKED ISSUE #834: Metadata - problem with editing metadata value
There is a problem when we want to replace the non-ascii text value in metadata with ascii one. In such a case we receive the error:
Failure_Metadata_Value_Size_Delta_Mismatch

Example steps:
1. Create metadata with non-ascii text, for example "秦"
2. Wait for transaction confirmation and try editing metadata to "t"

As a result, we receive the error: Failure_Metadata_Value_Size_Delta_Mismatch

Example payload:

0801000000000000A26BE365F3427509F0B65AC6F62EB87FA61D396B0770BEF18250CEA98E941FD7D2BF519B2E0A8B3C964F6AA746102A8B460E5BFEADEECA059A0530A373FB9208022CB0362389D0F0BEE5F078548AF7C4F63CA64A681B85BEE405C94D67D7F35500000000019841419096000000000000621B288901000000752EEBFB492FF329E7FBEE09865A208B08054C33731C857728EA757C6AF1546C60000000000000005A00000000000000022CB0362389D0F0BEE5F078548AF7C4F63CA64A681B85BEE405C94D67D7F355000000000198444198AA9CECDDF1A6C0657E34D8B5F7935C6A9383DFC739B0D5E35080E3FC91018EFEFF0600**C293C2A7C2A6**000000000000

The problem is related to how xor byte value is converted back to text:
https://github.com/symbol/symbol-sdk-typescript-javascript/blob/81f5397b6f2db1d1c3eea15d3e8a41b6b82ae31f/src/service/MetadataTransactionService.ts#L77

yilmazbahadir:
fixes #834 

### What was the issue?
from #834 
> There is a problem when we want to replace the non-ascii text value in metadata with ascii one. In such a case we receive the error:
Failure_Metadata_Value_Size_Delta_Mismatch

### What's the fix?
We noticed that `value` field in the Metadata Transactions(`AccountMetadataTransaction`, `NamespaceMetadataTransaction`, `MosaicMetadataTransaction`) has the type `string`.
This was the root cause of the issue, since the `value` field was meant to be a difference between the previous and the new value in _bytes_, converting the difference to first `string` and then to `byte[]` was resulting in loss of bytes and therefore sending wrong information to the `catapult-rest`.

In order to fix this, we:
* Removed the `rstr2utf8` method which was used during the conversion, fixing _insert_ part of the problem.
* Changed the `value` type from `string` to `Uint8Array`
* Added test cases for the non-ascii utf8 text <-> byte array conversion.

### DIFF
--- a/src/model/transaction/MosaicMetadataTransaction.ts
+++ b/src/model/transaction/MosaicMetadataTransaction.ts
@@ -51,8 +51,7 @@ export class MosaicMetadataTransaction extends Transaction {
      * @param scopedMetadataKey - Metadata key scoped to source, target and type.
      * @param targetMosaicId - Target unresolved mosaic identifier.
      * @param valueSizeDelta - Change in value size in bytes.
-     * @param value - String value with UTF-8 encoding
-     *                Difference between the previous value and new value.
+     * @param value - Difference between the previous value and new value.
      *                You can calculate value as xor(previous-value, new-value).
      *                If there is no previous value, use directly the new value.
      * @param maxFee - (Optional) Max fee defined by the sender
@@ -66,7 +65,7 @@ export class MosaicMetadataTransaction extends Transaction {
         scopedMetadataKey: UInt64,
         targetMosaicId: UnresolvedMosaicId,
         valueSizeDelta: number,
-        value: string,
+        value: Uint8Array,
         networkType: NetworkType,
         maxFee: UInt64 = new UInt64([0, 0]),
         signature?: string,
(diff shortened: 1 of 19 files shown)


src/model/transaction/MosaicMetadataTransaction.ts
```diff
@@ -51,8 +51,7 @@ export class MosaicMetadataTransaction extends Transaction {
      * @param scopedMetadataKey - Metadata key scoped to source, target and type.
```
gimre-xymcity, line 51: isn't this change breaking a existing code using typescript sdk?

yilmazbahadir: > isn't this change breaking a existing code using typescript sdk?

Yes, it is. 
I tried to keep the implications minimal. Most probably in the clients how a metadata tx is created won't change if they use https://github.com/symbol/symbol-sdk-typescript-javascript/blob/079297ddc765c61a058b0d132afbe8c8dd826c6f/src/service/MetadataTransactionService.ts#L60 as we do in our clients(desktop-wallet, symbol-cli).
But the value type used for presentation will change ofc.

Jaguar0625 (approved):

PR #835, opened 2022-02-15
fix: metadata value ser/deser is fixed

yilmazbahadir:
### What was the issue?
After the type of `Metadata.value` is changed from `string` to `byte array`,  we noticed that JSON serialization and deserialization started to fail.

### The fix
`byte array` is now being converted to `hex` value during serialization and converted back during deserialization.

### DIFF
--- a/src/core/crypto/Utilities.ts
+++ b/src/core/crypto/Utilities.ts
@@ -15,7 +15,6 @@
  */
 
 import * as CryptoJS from 'crypto-js';
-import { WordArray } from 'crypto-js';
 import * as hkdf from 'futoin-hkdf';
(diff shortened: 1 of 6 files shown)


Jaguar0625:
should there be regression tests for this change?

yilmazbahadir:
> should there be regression tests for this change?

I think there should, `SerializeTransactionToJSON.spec.ts` and `CreateTransactionFromDTO.spec.ts` doesn't cover the metadata txs so it couldn't catch this bug.  I will update the unit tests 👍

yilmazbahadir:
@Jaguar0625 Unit tests are added.

Jaguar0625 (approved):

Issue #849, opened 2023-02-26
The revocable attribute is missing when generating a MethodDefinitionTransaction from a payload.


xembook:
1. Expected behavior and actual behavior.

I want to create MethodDefinitionTransaction from payload,
but the revocable attribute is missing

```js
const mosaicDefTx = MosaicDefinitionTransaction.create( ... );
  console.log(mosaicDefTx);
  const serializedTx = mosaicDefTx.serialize();
  const deserializedTx = MosaicDefinitionTransaction.createFromPayload(serializedTx);
  console.log(deserializedTx);
```

This may be due to the fact that Revocable is not mentioned here.
https://github.com/symbol/symbol-sdk-typescript-javascript/blob/f094ab4be762986c593b0dbb1cb9cd4791977a33/src/model/transaction/MosaicDefinitionTransaction.ts#L155

Thank you for your consideration.

cryptoBeliever:
Fixed and released in 2023 (https://github.com/symbol/symbol-sdk-typescript-javascript/pull/850) - wasn't closed for some reason.

PR #850, opened 2023-02-26
Fix mosaic definition transaction from payload method

cryptoBeliever:
### What was the issue?

The revocable attribute is missing when generating a MethodDefinitionTransaction from a payload. Check #849

### What's the fix?

MethodDefinitionTransaction is initializing now all mosaic flags when creating MosaicDefinitionTransaction from the payload.


Jaguar0625 (approved):
👍

PR #853, opened 2024-02-12
fix: raw message encoding in TransferTransaction

OlegMakarenko:
## Problem
The message is not correctly encoded when creating a `TransferTransaction` with `RawMessage`.
The `Uint8Array` is encoded to a hex string during `RawMessage.create()`.
Then in the `getMessageBuffer()` of `TransferTransaction` the hex string (considered as a plain utf-8 text string) is converted to hex string again.
Also, an extra 0xFF byte is added at the beginning of the raw message.

## Solution 
- Remove double conversion of `RawMessage.payload` to hex in `TransferTransaction.getMessageBuffer()`, since the payload is already in hex.
- Remove adding an extra byte to the beginning of the message.

### DIFF
--- a/src/model/transaction/TransferTransaction.ts
+++ b/src/model/transaction/TransferTransaction.ts
@@ -202,13 +202,15 @@ export class TransferTransaction extends Transaction {
         if (!this.message || !this.message.payload) {
             return Uint8Array.of();
         }
-        const messgeHex =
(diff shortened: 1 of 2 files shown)


Jaguar0625 (approved):
👍

Wayonb (approved):

